PR #2, opened 2024-07-17
ValKey JSON module RFC

joehu21:
The proposed ValKey JSON module supports the native JavaScript Object Notation (JSON) format to encode complex datasets inside OSS ValKey. It is compliant with [RFC7159](http://www.ietf.org/rfc/rfc7159.txt) and [ECMA-404](http://www.ecma-international.org/publications/standards/Ecma-404.htm) JSON data interchange standard. With this feature, users can natively store, query, and modify JSON data structures in ValKey using the popular [JSONPath query language](https://www.ietf.org/archive/id/draft-goessner-dispatch-jsonpath-00.html). The module should be API-compatible and RDB-compatible with Redis Ltd.’s RedisJSON v2.


bbarani (approved):

JSON.md
```diff
@@ -0,0 +1,200 @@
...
+
+## Design Considerations
+
+ValKey JSON module introduces a new JSON data type for ValKey, and commands to insert, update, delete and query JSON data. 
+The module should be API-compatible and RDB-compatible with Redis Ltd.’s RedisJSON v2. Users can natively store, query, 
+and modify JSON data structures, either wholly or partially, using the popular [JSONPath query language](https://www.ietf.org/archive/id/draft-goessner-dispatch-jsonpath-00.html).
+
+![](./images/json_overview.png)
+
+### Module OnLoad
```
zuiderkwast, line 33: This section and the following few sections (including Keyspace Event Notification) simply specify the behaviour without any discussion about why, etc. It's better to move it to Specification.

zuiderkwast (commented):
Thanks!

The content looks very good in general, but I think some should be moved from design considerations to specification. I don't know if others agree. The RFC process is new for all of us.

zuiderkwast:
One more thing: Valkey is spelled with lowercase "k". 😄

JSON.md
```diff
@@ -0,0 +1,200 @@
...
+RFC: 2
+Status: Proposed
+---
+
+# ValKey JSON Module RFC
+
+## Abstract
+
+The proposed ValKey JSON module supports the native JavaScript Object Notation (JSON) format to encode complex datasets inside 
+OSS ValKey. It is compliant with [RFC7159](http://www.ietf.org/rfc/rfc7159.txt) and [ECMA-404](http://www.ecma-international.org/publications/standards/Ecma-404.htm)
```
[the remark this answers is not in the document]
joehu21: Dropped OSS.

JSON.md
```diff
@@ -0,0 +1,200 @@
...
+structures by providing powerful searching and filtering capabilities. However, [ValKey core](https://github.com/valkey-io/valkey) 
+does not have a native data type for JSON. Redis Ltd.‘s RedisJSON is a popular Redis modules, but not under a true 
+open source license and hence cannot be distributed freely with ValKey. There’s a demand in the ValKey 
+community to have a JSON module that matches most of the features of RedisJSON and is as API-compatible as possible. 
+See the community discussions [here](https://github.com/orgs/valkey-io/discussions?discussions_q=is%3Aopen+JSON).
+
+## Design Considerations
+
+ValKey JSON module introduces a new JSON data type for ValKey, and commands to insert, update, delete and query JSON data. 
+The module should be API-compatible and RDB-compatible with Redis Ltd.’s RedisJSON v2. Users can natively store, query,
```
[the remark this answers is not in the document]
joehu21: Added "why".

JSON.md
joehu21: Moved to Specification.

joehu21:
Updated all occurrences of "ValKey" to "Valkey", including the work in graphs.

zuiderkwast (approved):
LGTM

zuiderkwast:
@madolson do you want to take a quick look before we merge this?

ValkeyJSON.md
```diff
@@ -0,0 +1,214 @@
...
+rest, RapidJSON stands out as both memory efficient and providing efficient insert/update/delete API.
+
+### JSONPath Query Parser
+
+[JSONPath](https://www.ietf.org/archive/id/draft-goessner-dispatch-jsonpath-00.html) is a query language for JSON with 
+features similar to XPath for XML. JSONPath is used for selecting and extracting elements from a JSON document. It 
+supports advanced query capabilities such as wildcard, filter expressions, slices, union, recursive search, etc.
+Using JSONPath query language, users can natively store, query, and modify JSON data structures, either wholly or partially.
+
+![](./images/jsonpath_parser_design.png)
```
madolson, line 51: Can we move this to be a table inline with the text instead of a png? If anyone wants to comment or modify it, it will be difficult.

ValkeyJSON.md
```diff
@@ -0,0 +1,214 @@
...
+### ACL
+
+ValkeyJSON introduces a new ACL category - @json. The category has 22 JSON commands. No existing Valkey commands 
+are members of the @json category. 
+
+There are 4 existing ACL categories which are updated to include new JSON commands: @read, @write, @fast, @slow. The 
+table below has a column for each of these categories and a row for each command. If the cell contains a “y” then that 
+command must be added into that category. All of the other command members of those categories remain unchanged.
+
+![](./images/json_command_acl.png)
```
madolson, line 127: Ideally this should also be a table.

ValkeyJSON.md
```diff
@@ -0,0 +1,214 @@
...
+
+In earlier RFC 4627, only objects or arrays were allowed as root values of JSON. Since [RFC 7159](http://www.ietf.org/rfc/rfc7159.txt), 
+the root value of a JSON document can be of any type, scalar type (String, Number, Boolean, Null) or container type (Array, Object). 
+ValkeyJSON is compliant with [RFC 7159](http://www.ietf.org/rfc/rfc7159.txt).
+
+### JSON Command API
+
+There are 22 JSON commands. See https://docs.aws.amazon.com/memorydb/latest/devguide/json-list-commands.html. The API is 
+compatible with RedisJSON v2. Note that API compatibility here means our command API is a superset of RedisJSON API. 
+For example, we have command “JSON.DEBUG DEPTH” and “JSON.DEBUG FIELDS”, while they do not.
```
madolson, line 103: Ideally I would list them all out here. I don't think it makes a lot of sense to reference the AWS documentation for the set of commands.

madolson (commented):
@zuiderkwast since it becomes harder to comment after it's merged. I thought we should let everyone comment first before merging it as a proposal.

zuiderkwast:
> @zuiderkwast since it becomes harder to comment after it's merged. I thought we should let everyone comment first before merging it as a proposal.

Sounds reasonable. If this is the process we want, we should probably clarify that in the README at some point. The README says we'll merge the PR once the style and format is fine.

ValkeyJSON.md
```diff
@@ -0,0 +1,214 @@
...
+does not have a native data type for JSON. Redis Ltd.‘s RedisJSON is a popular Redis module, but not under a true 
+open source license and hence cannot be distributed freely with Valkey. There’s a demand in the Valkey 
+community to have a JSON module that matches most of the features of RedisJSON and is as API-compatible as possible. 
+See the community discussions [here](https://github.com/orgs/valkey-io/discussions?discussions_q=is%3Aopen+JSON).
+
+## Design Considerations
+
+ValkeyJSON will introduce a new JSON data type for Valkey, and commands to insert, update, delete and query JSON data. 
+To help users migrate from Redis and RedisJSON, as well as capitalize on existing OSS RedisJSON client libraries, ValkeyJSON 
+aims to be a drop-in replacement of RedisJSON. Therefore, it is designed be API-compatible and RDB-compatible with
```
daniel-house, line 30:
```suggestion
aims to be a drop-in replacement of RedisJSON. Therefore, it is designed to be API-compatible and RDB-compatible with 
```

ValkeyJSON.md
```diff
@@ -0,0 +1,214 @@
...
+features similar to XPath for XML. JSONPath is used for selecting and extracting elements from a JSON document. It 
+supports advanced query capabilities such as wildcard, filter expressions, slices, union, recursive search, etc.
+Using JSONPath query language, users can natively store, query, and modify JSON data structures, either wholly or partially.
+
+![](./images/jsonpath_parser_design.png)
+
+One of the deficiencies of the RapidJSON library is that it does not support JSONPath query. RapidJSON only supports 
+[JSON Pointer](https://datatracker.ietf.org/doc/html/rfc6901) query language, which is restricted to single-value selection. 
+Single-value selection corresponds to the restricted query capability in RedisJSON v1 API. RedisJSON v2 API supports a 
+much powerful JSONPath syntax with wildcard, filter expressions, slices, union, recursive search, etc., and can select
```
daniel-house, line 56:
```suggestion
much more powerful JSONPath syntax with wildcard, filter expressions, slices, union, recursive search, etc., and can select 
```

ValkeyJSON.md
```diff
@@ -0,0 +1,214 @@
...
+![](./images/jsonpath_parser_design.png)
+
+One of the deficiencies of the RapidJSON library is that it does not support JSONPath query. RapidJSON only supports 
+[JSON Pointer](https://datatracker.ietf.org/doc/html/rfc6901) query language, which is restricted to single-value selection. 
+Single-value selection corresponds to the restricted query capability in RedisJSON v1 API. RedisJSON v2 API supports a 
+much powerful JSONPath syntax with wildcard, filter expressions, slices, union, recursive search, etc., and can select 
+multiple JSON values. RedisJSON v2 API is a superset of v1 API.
+
+We should extend RapidJSON to support a subset of the JSONPath query language that is compatible with RedisJSON v2 API. 
+To achieve this, we designed a JSONPath query parser that integrates with RapidJSON. It is a recursive descent parser
```
daniel-house, line 60: From here until the end of this section as well as the diagram above appear to be design details that over-commit to a specific solution.

For example, why should it be a recursive descent parser?  For me, recursive descent implies a hand written parser: why not a generated via ANTLR or even yacc?

ValkeyJSON.md
```diff
@@ -0,0 +1,214 @@
...
+and commit, where prepareSetValues does not change the Valkey data. The purpose of having a 2-phase write is to be able 
+to discard the write operation if certain conditions are not satisfied.
+
+### Tokenization of JSON Object Keys
+
+JSON has two container types - Array and Object. The JSON Object is a key/value mapping where keys are restricted to 
+be strings, whereas the value can be any valid JSON value, including another container value. Thus, a single JSON document 
+can contain multiple JSON objects, each with its own unique namespace. It’s quite common that a key repeatedly appears 
+in the same JSON document or across documents. If the number of JSON documents is large, repeated key names can consume 
+significant amount of memory usage.
```
daniel-house, line 81:
```suggestion
a significant amount of memory usage.
```

ValkeyJSON.md
```diff
@@ -0,0 +1,214 @@
...
+
+To remove duplicate copies of JSON object key names, the module tokenizes key names by maintaining a global data structure 
+called KeyTable, which is a sharded hash table storing key tokens and reference counts. Access to the KeyTable is threadsafe.
+
+## Specification
+
+### Supported JSON Standard
+
+[RFC7159](http://www.ietf.org/rfc/rfc7159.txt) and [ECMA-404](http://www.ecma-international.org/publications/standards/Ecma-404.htm)
```
daniel-house, line 90:
```suggestion
[RFC7159](http://www.ietf.org/rfc/rfc7159.txt) and [ECMA-404](https://ecma-international.org/publications-and-standards/standards/ecma-404/) 
```

The original link was not found.

ValkeyJSON.md
```diff
@@ -0,0 +1,214 @@
...
+### Supported JSON Standard
+
+[RFC7159](http://www.ietf.org/rfc/rfc7159.txt) and [ECMA-404](http://www.ecma-international.org/publications/standards/Ecma-404.htm) 
+JSON data interchange standard is supported. UTF-8 Unicode in JSON text is supported.
+
+### Root JSON Value
+
+In earlier RFC 4627, only objects or arrays were allowed as root values of JSON. Since [RFC 7159](http://www.ietf.org/rfc/rfc7159.txt), 
+the root value of a JSON document can be of any type, scalar type (String, Number, Boolean, Null) or container type (Array, Object). 
+ValkeyJSON is compliant with [RFC 7159](http://www.ietf.org/rfc/rfc7159.txt).
```
daniel-house, line 97:
```suggestion
ValkeyJSON MUST BE compliant with [RFC 7159](http://www.ietf.org/rfc/rfc7159.txt).
```

I think this is more in line with RFC conventions.

ValkeyJSON.md
daniel-house: It would also help to call out and give details about _all_ the commands that we add.

ValkeyJSON.md
```diff
@@ -0,0 +1,214 @@
...
+Two kinds of JSONPath query syntaxes are supported:
+* Enhanced syntax – Follows the [Goessner-style](https://goessner.net/articles/JsonPath/) JSONPath query syntax, as shown in the table below.
+* Restricted syntax – Has limited query capabilities, compatible with RedisJSON v1.
+
+If a query path starts with '$', the enhanced syntax is used. Otherwise, the restricted syntax is used. A query using 
+the enhanced syntax always returns an array of values, while a restricted-syntax query always returns a single value. 
+
+![](./images/jsonpath_syntax.png)
+
+For details and examples, see https://docs.aws.amazon.com/memorydb/latest/devguide/json-document-overview.html#json-path-syntax.
```
daniel-house, line 116: See earlier comments regarding PNGs and references to AWS documentation.

ValkeyJSON.md
```diff
@@ -0,0 +1,214 @@
...
+ValkeyJSON introduces a new ACL category - @json. The category has 22 JSON commands. No existing Valkey commands 
+are members of the @json category. 
+
+There are 4 existing ACL categories which are updated to include new JSON commands: @read, @write, @fast, @slow. The 
+table below has a column for each of these categories and a row for each command. If the cell contains a “y” then that 
+command must be added into that category. All of the other command members of those categories remain unchanged.
+
+![](./images/json_command_acl.png)
+
+### Info metrics
```
daniel-house, line 129:
```suggestion
### Info Metrics
```

ValkeyJSON.md
```diff
@@ -0,0 +1,214 @@
...
+
+Info metrics are visible through the “info json” or “info modules” command.
+
+| Name                    | 	Description                                                           |
+|:------------------------|:---------------------------------------------------------------------|
+| json_total_memory_bytes | 	Total amount of memory allocated to JSON documents and meta data. |
+| json_num_documents	  | Number of JSON keys                                                   |
```
daniel-house, line 136:
```suggestion
| json_num_documents	  | Number of JSON keys.                                                  | 
```

Punctuation style should be consistent.  This applies to other tables as well.  Instead of the suggested change, just stop using periods in all these tables.

ValkeyJSON.md
```diff
@@ -0,0 +1,214 @@
...
+Info metrics are visible through the “info json” or “info modules” command.
+
+| Name                    | 	Description                                                           |
+|:------------------------|:---------------------------------------------------------------------|
+| json_total_memory_bytes | 	Total amount of memory allocated to JSON documents and meta data. |
+| json_num_documents	  | Number of JSON keys                                                   | 
+
+### Configs
+
+![](./images/json_configs.png)
```
daniel-house, line 140: See comments on PNGs elsewhere.

ValkeyJSON.md
```diff
@@ -0,0 +1,214 @@
...
+data structure, not the serialized JSON string. The value can be overridden by module config json.max-document-size.
+
+The amount of memory consumed by a JSON document can be inspected by 
+using the “JSON.DEBUG MEMORY" or “MEMORY USAGE” command. “JSON.DEBUG MEMORY <key> [<path>]" can also report the size 
+of a JSON sub-tree.
+
+### Nesting Depth Limit
+
+When a JSON object or array has an element that is itself another JSON object or array, that inner object or array is 
+said to “nest” within the outer object or array. To avoid stack overflow, it's good to have a limit on the nesting depth.
```
daniel-house, line 156: This a standard reason for not using naive recursive descent parsing.

ValkeyJSON.md
```diff
@@ -0,0 +1,214 @@
...
+The default path limit is 128 levels. Any attempt to create a document that contains a nesting depth greater than 128 
+will be rejected with an error message. The value can be overridden by module config json.max-path-limit.
+
+### Module API
+
+ValkeyJSON integrates with Valkey core by implementing Valkey module API.
```
daniel-house, line 162:
```suggestion
ValkeyJSON integrates with Valkey core via calls to the Valkey module API.
```

ValkeyJSON.md
```diff
@@ -0,0 +1,214 @@
...
+
+![](./images/json_overview.png)
```
daniel-house, line 164: I am uncomfortable with this diagram.  The replica seems to me to be superfluous to the information being communicated.  The `JSON Module` looks like it is part of Valkey when modules created via the Module API are optional and fully external to Valkey.

ValkeyJSON.md
joehu21: Will remove the diagram. Also, based on other feedbacks, will make the design description stay at high level.

ValkeyJSON.md
joehu21: Converted this and all other graphs to text.

ValkeyJSON.md
joehu21: Will document all commands

ValkeyJSON.md
joehu21: updated.

ValkeyJSON.md
joehu21: updated

ValkeyJSON.md
joehu21: Agreed it's too much design details. Reduced the description to stay at high level. BTW, I don't think recursive descent (RD) implies a hand-written parser. RD doesn't imply hand-written or code-gen. Many code-gen tools uses RD under the hood. A subset of LL(k) grammar can be implemented using RD.

ValkeyJSON.md
joehu21: updated

ValkeyJSON.md
joehu21: updated

ValkeyJSON.md
joehu21: "must be" is too strong. Changed to "will be". Yes, will use "should, will, must..." throughout the RFC.

ValkeyJSON.md
joehu21: Removed the PNGs and references AWS documentation. It's now all self-contained markdown text.

ValkeyJSON.md
joehu21: updated

ValkeyJSON.md
joehu21: updated

ValkeyJSON.md
joehu21: Will remove all PNGs. Configs can be changed dynamically.  Having a bound on object size and tree depth is generally considered as a good design practice. One use case is: A malicious program can repeatedly call "json.arrinsert" command to make a json key grow forever, leading to OOM. Another use case: A json tree without depth limit may lead to stack overflow if some code recursively traverse the tree.

ValkeyJSON.md
joehu21: Not exactly "via calls to the Valkey module API". Most of the integration is done through callbacks. Will rephrase it.

ValkeyJSON.md
joehu21: Removed the diagram.

ValkeyJSON.md
daniel-house: I share both of your concerns, malicious programs and stack overflows, but I feel there is a better solution.

Regarding a json structure that causes OOM.  This complaint applies equally to lists, hashes, sets, etc., and we don't set upper limits for them in the config file.  It would be more consistent with other Valkey data structures to catch this at the time the structure becomes too large and either reject the operation or enter an OOM state.  This could very will be caused by an attack, but also by a bug or simply be the correct behavior from the point of view of the system designer.  It seems a little strange, but it is entirely possible for a single set to occupy the entirety of Valkey, so why not a Json structure?  (As a sillier example, it is possible for the value to be a boolean while the key is a 512 megabyte string.  See [keyspace](https://valkey.io/topics/keyspace/))

Regarding recursive traversal causing a stack overflow.  Naturally, this would be unacceptable as it crashes the server.  Solving this via a configuration property feels wrong to me.  It feels like the correct value depends on too many things that are outside our control.  Recursive descent parsing is very easy to implement (by hand or by a generator) but should be avoided unless the grammar (unlike this one) has an inherently small maximum depth.  The same code can be rewritten in a reasonably clear and straightforward manner.  There have been many discussions of this over the years, here is one: [Replace Recursion With Iteration](https://www.refactoring.com/catalog/replaceRecursionWithIteration.html).  The first sentence is particularly amusing.

ValkeyJSON.md
daniel-house: The callbacks are registered via calls to the Valkey module API, yes?

Your new phrasing, "through Valkey module API callbacks or calls to the Valkey core" still troubles me.  The callback code is in the shared library, not the API.  Calls are made to the module API I hope, not the core.  (We do have at least one module that calls the core.  If the JSON module will call the core that is whole other can of worms.)  Perhaps something less specific like "ValkeyJSON shall be implemented via the [Valkey modules API](https://valkey.io/topics/modules-intro/)", or, even simpler, "... shall be a Valkey module, see ...".

joehu21: "ValkeyJSON shall be implemented via the [Valkey modules API](https://valkey.io/topics/modules-intro/)" sounds good to me.

ValkeyJSON.md
joehu21: Understood. Keep the text for now. Let's also see what others think.

bbarani:
@madolson @daniel-house @zuiderkwast Whats the next step here? Are we good to merge this PR?

ValkeyJSON.md
```diff
@@ -0,0 +1,894 @@
...
+| Symbol/Expression | Description                                                              |
+|:------------------|:-------------------------------------------------------------------------|
+| $                 | the root element                                                         |
+| . or []           | child operator                                                           |
+| ..                | recursive descent                                                        |
+| *                 | wildcard. All elements in an object or array.                            |
+| []                | array subscript operator. Index is 0-based.                              |
+| [,]               | union operator                                                           |
+| [start:end:step]  | array slice operator                                                     |
+| ?()               | applies a filter expression to the current array or object               |                                                      |
```
madolson, line 125:
```suggestion
| ?()               | applies a filter expression to the current array or object               |
```

ValkeyJSON.md
```diff
@@ -0,0 +1,894 @@
...
+| '$..book[?(@.price>=10 && @.price<=100)]'                     | all books in the price range of $10 to $100, inclusive. (The path must be quoted if it contains whitespaces) |
+| $..book[?(@.sold==true&#124;&#124;@.in-stock==false)]         | all books sold or out of stock                         |
+| '$..book[?(@.sold == true &#124;&#124; @.in-stock == false)]' | all books sold or out of stock. (The path must be quoted if it contains whitespaces)                 |
+| '$.store.book[?(@.["category"] == "fiction")]'                | all books in the fiction category |
+| '$.store.book[?(@.["category"] != "fiction")]'                | all books in non-fiction categories |
+
+### JSON Command API
+
+The API is compatible with RedisJSON v2. Note that API compatibility here means our command API is a superset of RedisJSON API. 
+For example, we have command “JSON.DEBUG DEPTH” and “JSON.DEBUG FIELDS”, while they do not.
```
madolson, line 168: I would like to know why not have two separate commands, JSON.DEPTH and JSON.FIELDs ? Having a debug command doesn't really make that much sense.

madolson:
This RFC looks good to me as well. As per our updated guidance, @valkey-io/core-team please take a look and review offline line. We'll schedule to review this next week in our team meeting: 2024-08-12. I'll send an invite to Joe.

ValkeyJSON.md
daniel-house: Thanks for adding the command specifications.  That is most helpful.

Which commands are in the Valkey superset of the RedisJSON API?  So far as I can tell, only JSON.DEBUG (with 5 supported subcommands).  Is that all?  Is it intended that all extensions be subcommands of JSON.DEBUG?

ValkeyJSON.md
joehu21: The intent is to avoid diverging from RedisJSON unnecessarily. If we introduce new commands, they won't be supported by the existing OSS RedisJSON client libraries. If we have our own ValkeyJSON client library ecosystem, then it would make sense to have them as new commands, instead of sub-commands of JSON.DEBUG.

joehu21: In the RFC, “JSON.DEBUG DEPTH” and “JSON.DEBUG FIELDS” are the only extension. We consider "depth of a json tree/sub-tree" and "number of fields in a json tree/sub-tree" as debug informations. Hence, they are subcommands of JSON.DEBUG. However, it doesn't mean in the future, all extensions should be subcommands of JSON.DEBUG. I am open to adding new JSON commands in the future if necessary.

daniel-house: I tried to look up the list of commands elsewhere and my best source for the RedisJSON v2 API didn't agree with this.  Probably I am not looking in the right place.  Please point to it for me.

I saw JSON.DEBUG without arguments (which is not in the PR), and JSON.DEBUG MEMORY (which is), but not JSON.DEBUG HELP (an extension?).  I also saw JSON.MSET (an oversight?).

madolson: > I also saw JSON.MSET (an oversight?)

Yeah, we missed this. This was added sort of randomly by Redis at some point.

joehu21: RedisJSON API is evolving. Missed JSON.MSET. Will add it to the RFC when I get a chance.

Interestingly, their documentation doesn't seem to match latest docker image `redis/redis-stack:latest`. I don't see I don't see json.mset:
```
docker run -p 6379:6379 redis/redis-stack:latest
...
% ./redis-cli command |grep -i json
json.arrtrim
json.arrpop
json.clear
json.arrappend
json.objlen
json.nummultby
json.mget
json.strappend
json.toggle
json.get
json.arrlen
json.strlen
json.numpowby
json.resp
json.type
json.forget
json.arrindex
json.arrinsert
json.set
json.del
json.numincrby
json.debug
json.objkeys
```

Also, using the same docker image, I only see `JSON.DEBUG HELP` and `JSON.DEBUG` is invalid.
```
127.0.0.1:6379> json.debug help
1) "MEMORY <key> [path] - reports memory usage"
2) "HELP                - this message"
127.0.0.1:6379> json.debug
(error) ERR wrong number of arguments for 'json.debug' command
```

joehu21: Just realized RedisJSON added `json.numpowby` at some point. Will add this to the RFC as well.

ValkeyJSON.md
```diff
@@ -0,0 +1,894 @@
...
+Valkey using the popular [JSONPath query language](https://www.ietf.org/archive/id/draft-goessner-dispatch-jsonpath-00.html). 
+To help users migrate from Redis and RedisJSON, as well as capitalize on existing OSS RedisJSON client libraries, the module 
+is designed to be API-compatible and RDB-compatible with Redis Ltd.’s RedisJSON v2.
+
+## Motivation
+
+JSON format is a widely used data exchange format and simplifies the development of applications that store complex data 
+structures by providing powerful searching and filtering capabilities. However, [Valkey core](https://github.com/valkey-io/valkey) 
+does not have a native data type for JSON. Redis Ltd.‘s RedisJSON is a popular Redis module, but not under a true 
+open source license and hence cannot be distributed freely with Valkey. There’s a demand in the Valkey
```
enjoy-binbin, line 22:
```suggestion
open source license and hence cannot be distributed freely with Valkey. There's a demand in the Valkey 
```

ValkeyJSON.md
```diff
@@ -0,0 +1,894 @@
...
+| JSON.NUMINCRBY | y     |       | y      | y    |       |
+| JSON.NUMMULTBY | y     |       | y      | y    |       |
+| JSON.OBJKEYS   | y     | y     |        | y    |       |
+| JSON.OBJLEN    | y     | y     |        | y    |       |
+| JSON.RESP      | y     | y     |        | y    |       |
+| JSON.SET       | y     |       | y      |      | y     |
+| JSON.STRAPPEND | y     |       | y      | y    |       |
+| JSON.STRLEN    | y     | y     |        | y    |       |
+| JSON.TOGGLE    | y     |       | y      | y    |       |
+| JSON.TYPE      | y     | y     |        | y    |       |
```
enjoy-binbin, line 824:
```suggestion
| JSON Command   | @json | @read | @write | @fast | @slow |
|:---------------|:------|:------|:-------|:------|:------|
| JSON.ARRAPPEND | y     |       | y      | y     |       |
| JSON.ARRINDEX  | y     | y     |        | y     |       |
| JSON.ARRINSERT | y     |       | y      | y     |       |
| JSON.ARRLEN    | y     | y     |        | y     |       |
| JSON.ARRPOP    | y     |       | y      | y     |       |
| JSON.ARRTRIM   | y     |       | y      | y     |       |
| JSON.CLEAR     | y     |       | y      | y     |       |
| JSON.DEBUG     | y     | y     |        |       | y     |
| JSON.DEL       | y     |       | y      | y     |       |
| JSON.FORGET    | y     |       | y      | y     |       |
| JSON.GET       | y     | y     |        | y     |       |
| JSON.MGET      | y     | y     |        | y     |       |
| JSON.NUMINCRBY | y     |       | y      | y     |       |
| JSON.NUMMULTBY | y     |       | y      | y     |       |
| JSON.OBJKEYS   | y     | y     |        | y     |       |
| JSON.OBJLEN    | y     | y     |        | y     |       |
| JSON.RESP      | y     | y     |        | y     |       |
| JSON.SET       | y     |       | y      |       | y     |
| JSON.STRAPPEND | y     |       | y      | y     |       |
| JSON.STRLEN    | y     | y     |        | y     |       |
| JSON.TOGGLE    | y     |       | y      | y     |       |
| JSON.TYPE      | y     | y     |        | y     |       |
```

ValkeyJSON.md
```diff
@@ -0,0 +1,894 @@
...
+
+### Info Metrics
+
+Info metrics are visible through the “info json” or “info modules” command.
+
+| Info Name               | 	Description                                                    |
```
enjoy-binbin, line 830:
```suggestion
| Info Name               | 	Description                                                   |
```

ValkeyJSON.md
```diff
@@ -0,0 +1,894 @@
...
+|:------------------------|:------------------------------------------------------------------|
+| json_total_memory_bytes | Total amount of memory allocated to JSON documents and meta data. |
+| json_num_documents      | Number of JSON keys.                                              | 
+
+### Module Configs
+
+| Config Name            | Default Value | Unit | 	Description                                          |
```
enjoy-binbin, line 837:
```suggestion
| Config Name            | Default Value | Unit | 	Description                                         |
```

ValkeyJSON.md
```diff
@@ -0,0 +1,894 @@
...
+
+The query expression should be designed to work with selecting a vector of values instead of a single value, and 
+compatible with both RedisJSON v1 and v2 query syntax. The query parser should automatically detect if the query is of 
+v1 or v2 syntax.
+
+### Tokenization of JSON Object Keys
+
+JSON has two container types - Array and Object. The JSON Object is a key/value mapping where keys are restricted to 
+be strings, whereas the value can be any valid JSON value, including another container value. Thus, a single JSON document 
+can contain multiple JSON objects, each with its own unique namespace. It’s quite common that a key repeatedly appears
```
enjoy-binbin, line 69:
```suggestion
can contain multiple JSON objects, each with its own unique namespace. It's quite common that a key repeatedly appears 
```

ValkeyJSON.md
```diff
@@ -0,0 +1,894 @@
...
+
+Upon loading, the module registers a new JSON data type. All operations such as query, insert, update and delete are 
+efficiently performed on the in-memory document objects, as opposed to JSON text.
+
+* Module name: json
+* JSON data type name: ReJSON-RL (Note: We use the same name as the one in RedisJSON for the sake of RDB compatibility.)
+
+#### Persistence
+
+ValkeyJSON hooks into Valkey’s persistence API via the module type callbacks:
```
enjoy-binbin, line 856:
```suggestion
ValkeyJSON hooks into Valkey's persistence API via the module type callbacks:
```

enjoy-binbin (approved):
i did not review all the lines carefully, LGTM with a rough look.

ValkeyJSON.md
joehu21: What's the suggested change? I don't see any difference.

ValkeyJSON.md
joehu21: What's the suggested change? I don't see any difference.

ValkeyJSON.md
```diff
@@ -0,0 +1,914 @@
...
+| Info Name               | 	Description                                                    |
+|:------------------------|:------------------------------------------------------------------|
+| json_total_memory_bytes | Total amount of memory allocated to JSON documents and meta data. |
+| json_num_documents      | Number of JSON keys.                                              | 
+
+### Module Configs
+
+| Config Name            | Default Value | Unit | 	Description                                          |
+|:-----------------------|:--------------|:-----|:------------------------------------------------------|
+| json.max-document-size | 64            | MB   | Maximum memory allowed for a single JSON document.    |
```
madolson, line 859: Please add an open question about what we want to do about this config. We want to be compatible with RedisJSON. Either:
1. Make the default value 0.
2. Remove the config until users ask for it.

joehu21: added the open question.

ValkeyJSON.md
```diff
@@ -859,6 +866,13 @@ Info metrics are visible through the “info json” or “info modules” comma
 | json.max-document-size | 64            | MB   | Maximum memory allowed for a single JSON document.    |
 | josn.max-path-limit    | 128           |      | Maximum nesting levels within a single JSON document. |
 
+TBD:
+
+We are considering a review feedback of not enforcing document size limit, because RedisJSON doesn't have the limit and Valkey 
+core data types don't have limits either. There could be two options:
+1. Make the default value of json.max-document-size 0 or -1, meaning unlimited.
+2. Remove the config.
```
madolson, line 874: Can you also move this out of the current section, it's supposed to be its own section in the doc.

```suggestion:

## Open Questions

We are considering a review feedback of not enforcing document size limit, because RedisJSON doesn't have the limit and Valkey 
core data types don't have limits either. There could be two options:
1. Make the default value of json.max-document-size 0 or -1, meaning unlimited.
2. Remove the config.
```

ValkeyJSON.md
madolson: Can you move it to its own section of its doc?

ValkeyJSON.md
joehu21: updated

madolson (approved):

ValkeyJSON.md
madolson: It's a spacing issue.

ValkeyJSON.md
```diff
@@ -0,0 +1,922 @@
...
+Info metrics are visible through the “info json” or “info modules” command.
+
+| Info Name               | 	Description                                                   |
+|:------------------------|:------------------------------------------------------------------|
+| json_total_memory_bytes | Total amount of memory allocated to JSON documents and meta data. |
+| json_num_documents      | Number of JSON keys.                                              | 
+
+### Module Configs
+
+| Config Name            | Default Value | Unit | 	Description                                         |
```
madolson, line 857:
```suggestion
| Config Name            | Default Value | Unit | 	Description                                     |
```

ValkeyJSON.md
madolson: I have a feeling it might be deprecated. I found out the json.nummultiply is deprecated as per     https://github.com/redis/redis-py/blob/1cbb97ecd7a0bf87aec332a257b589919385d90d/redis/commands/json/commands.py#L144C6-L144C25.

PR #3, opened 2024-07-23
Introduce admin port parameter to differentiate the admin client connection and general client connection RFC

hwware:
I would like to introduce another parameter in valkey.conf:  admin-port

Background:

The parameter "maxclient" default value is 10000. Client administrators could set higher value to accept more client connection.
However, there is one situation: If the connection number of client reaches the maxclient, then even administrators have no chance to connect to the server to process some urgent issues or update some configurations.

To solve above case, by connecting to the admin port, an administrator or cloud provider has the chance to connect to the server even the max number of connected clients reaches

### DIFF
--- a/ADMIN-PORT.md
+++ b/ADMIN-PORT.md
@@ -17,7 +17,8 @@ We want to let adminstrator connect to this admin port to do some special things
 
 Beside above mentioned, we could implement the following features in the future:
 1. The connections through admin port are not limited to maxclient number, we can always guarantee they can connect to the server.
-2. Some status check commands could be sent and receied through admin port, we can do some special logic to make sure they have opportunity to be handle first
+2. Some status check commands could be sent and receied through admin port, we can do some special logic to make sure they have opportunity to be handle first, such as:
+   PING, HELLO, ECHO, INFO etc
 
 
 ## Design Considerations
--- a/ADMIN-PORT.md
+++ b/ADMIN-PORT.md
@@ -37,6 +37,48 @@ Based on all above 3 candidate defects, we decide to select "admin-port" to impl
 4. If a client connects via admin-port and ACL rules apply to this client as well, this client should follow the ACL to execute some commands
 5. If a client connects via non-admin-port and ACL rules apply to this client as well, this client is not allowed to execute those special commands. The special commands could be new commands introduced by admin through modules, for example, setting/getting traffic/replication control threshold, encrypting the username and password in config file, etc
 
+## Sample for Module
+Assume we have one module, which only allows the administrator to enable the traffic control command: valkey.traffic.control.enable
+
+step 1: add a new variable in struct connection as: 
+        struct connection {
+            .........
+            int connected_port;
+        }
+
+step 2: create a function getClientConnectedPort(client *c) to get the connected server port for this specific client:
+        int getClientConnectedPort(client *c) {
+            .........
+            return c->conn->connected_port;
+        } 
+
+step 3: create an API in module:
+        int VM_GetClientConnectedPort(ValkeyModuleCtx *ctx) {
+            return getClientConnectedPort(ctx->client);
+        }
+        Thus, module developers can call it to get the client connected server port.
+
+step 4: In the module, we could have the following example: (Here, we assume the admin port is 7001)
+        int valkey_traffic_control_enable(ValkeyModuleCtx *ctx, ValkeyModuleString **argv, int argc) {
+            VALKEYMODULE_NOT_USED(argv);
+            VALKEYMODULE_NOT_USED(argc);
+            int port = ValkeyModule_GetClientConnectedPort(ctx);
+            if (port == 7001) {
+                ValkeyModule_ReplyWithSimpleString(ctx, "You can enable traffic control feature");
+            } else {
+                ValkeyModule_ReplyWithSimpleString(ctx, "You have no permission to execute this command");
+            }
+            return VALKEYMODULE_OK;
+        }
+
+        int ValkeyModule_OnLoad(ValkeyModuleCtx *ctx, ValkeyModuleString **argv, int argc) {
+            ..........
+            if (ValkeyModule_CreateCommand(ctx, "valkey.traffic.control.enable",
+                                                 valkey_traffic_control_enable, "", 0, 0,
+                                                 0) == VALKEYMODULE_ERR)
+            ..........
+        } 
+
 ## References
 Add a management-port https://github.com/valkey-io/valkey/issues/497
 Trusted/un-trusted client feature https://github.com/valkey-io/valkey/issues/469, https://github.com/valkey-io/valkey/pull/666
--- a/ADMIN-PORT.md
+++ b/ADMIN-PORT.md
@@ -9,23 +9,20 @@ Status: Proposed
 
 In Valkey.conf, Add a new config parameter: "admin-port" to allow cloud providers or internal administrators in a company or special users from public internet to connect to the Valkey server to execute some special commands or load specfic modules.
 
-
 ## Motivation
 
-For adminstrators, they may want to execute some special commands in the server, which are usually defined in the module, and
-other general clients are not allowed to run these kinds of commands. Thus connecting to an admin-port is a good way to implement this feature. 
+For adminstrators, they may want to execute some special commands for system administration in the server, such as authentication or traffic control commands, which are usually defined in the module. And other general clients are not allowed to run these kinds of commands. Thus connecting to an admin-port is a good way to implement this feature. 
 
-Beside above mentioned, we could implement the following features in the future once admin-port is commited:
+Beside above mentioned, we could implement the following enhancement features in the future once admin-port is commited:
 1. The connections through admin port are not limited to maxclient number, we can always guarantee the administrators can connect to the server.
-2. Some status check commands could be sent and receied through admin port, we can do some special logic to make sure they have opportunity to be handle first, such as: PING, HELLO, ECHO, INFO etc
-
+2. Some status check commands or dangerous commands could be sent and received through admin port, we can do some special logic to make sure they have opportunity to be handle first, such as: PING, HELLO, ECHO, INFO, FLUSHALL etc
 
 ## Design Considerations
 
 Before we propose the "admin-port" feature, we have the following 3 candidates:
-1. ACL: Now Valkey have ACL feature to implement similar function, but it brings one disadvantages: username need to be predefined.
+1. ACL: Now Valkey have ACL feature to implement similar function, but it brings two disadvantages: firstly, username need to be predefined and secondly, total number of client connections can not exceed the maxclient. Once the emergence issue happens and maxclient number reaches, the privilege user can not connect to the server any more.
 
-2. Run some special commands through the connected IP address: Before we run every commands, we need check if the client ip address is in a linkedlist, it costs CPU and extra memory, it causes side effect in performance
+2. Run some special customize commands for system administration through the connected IP address: Before we run every commands, we need check if the client ip address is in a linkedlist, it costs CPU and extra memory, it causes side effect in performance
 
 3. Iptables: It works on transport layer and IP address, but it can not apply to the specific commands of the Valkey
 
@@ -35,8 +32,10 @@ Based on all above 3 candidate defects, we decide to select "admin-port" to impl
 ## Specification
 
 1. The admin client role is only decided by the cloud providers or internal administrators, whatever where the client comes from. Thus generally, admin-port should not be exposed to the public.
-2. If a client connects via admin-port and ACL rules apply to this client as well, this client should follow the ACL to execute some commands
-3. If a client connects via non-admin-port and ACL rules apply to this client as well, this client is not allowed to execute those special commands. The special commands could be new commands introduced by admin through modules, for example, setting/getting traffic/replication control threshold, encrypting the username and password in config file, etc
+2. For every module, there is the API RedisModule_CreateCommand which defines a function to execute your customize command. You can find a way to decide if the command can be called by the client.
+3. For those Valkey built-in commands, generally we do not consider Valkey built-in commands as the special commands. Of course, now we can consider FLUSHALL command as an exception.
+4. If a client connects via admin-port and ACL rules apply to this client as well, this client should follow the ACL to execute some commands
+5. If a client connects via non-admin-port and ACL rules apply to this client as well, this client is not allowed to execute those special commands. The special commands could be new commands introduced by admin through modules, for example, setting/getting traffic/replication control threshold, encrypting the username and password in config file, etc
 
 ## References
 Add a management-port https://github.com/valkey-io/valkey/issues/497
--- a/ADMIN-PORT.md
+++ b/ADMIN-PORT.md
@@ -1,0 +1,52 @@
+---
+RFC: 1
+Status: Informational
+---
+
+Valkey RFC
+==========
+
+This repository is a collection of feature proposals and descriptions of changes to Valkey
+that require some more detail than just the text in a pull request or an issue.
+It is loosely inspired by RFCs and by Python's enhancement proposals (PEP).
+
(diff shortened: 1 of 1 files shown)


ADMIN-PORT.md
```diff
@@ -0,0 +1,12 @@
+ADMIN PORT
+==========
+Reference the issue https://github.com/valkey-io/valkey/issues/497 (add a management-port) and https://github.com/valkey-io/valkey/issues/469 (Trusted/un-trusted client feature)
+
+Background: In some cases, we can run valkey server on a container and multiply containers could be run in a physical machine simultaneously.
+            Adminstator wants to execute some special commands in the server, and general clients are not allowed to run these kinds of commands.
+            Thus, we want to set an admin port for adminstator connection, and general clients could still connect with server with port argument.
+
+            Although Valkey invloved ACL concept, but we need to set user first. But for this case, we have no idea the admin clients name and general
```
zuiderkwast, line 9: This is rendered as a code block. Did you check how the page is rendered using "View file" under the "..." menu?

zuiderkwast (commented):
This RFC contains much less than the issues. The idea is that the RFC should be more detailed than issues.

Please include the discussions of both issues, comparison of both approaches any why each decision is made.

Also please try to structure it a bit more like described in the README.

ADMIN-PORT.md
```diff
@@ -0,0 +1,44 @@
+---
+RFC: 3
+Status: Proposed
+---
+
+# ADMIN PORT RFC
+
+## Abstract
+
+In Valkey.conf, Add a new config parameter: "admin-port" to allow internal administrators in a company or special users from public internet to connect to the Valkey server to execute some special commands or load specfic modules.
```
madolson, line 10: Would we add it to all of the debug configs: https://github.com/valkey-io/valkey/blob/unstable/valkey.conf#L130 ?

ADMIN-PORT.md
```diff
@@ -0,0 +1,44 @@
...
+
+
+## Motivation
+
+For adminstrators, they may want to execute some special commands in the server, and other general clients are not allowed to run these kinds of commands.
+We want to let adminstrator connect to this admin port to do some special things. 
+
+Beside above mentioned, we could implement the following features in the future:
+1. The connections through admin port are not limited to maxclient number, we can always guarantee they can connect to the server.
```
madolson, line 19: I thought this was one of the main features? It seems like a good idea to me. I guess maybe instead of unlimited, maybe there is a special allowance of like 128 "admin" connections.

ADMIN-PORT.md
```diff
@@ -0,0 +1,44 @@
...
+2. Some status check commands could be sent and receied through admin port, we can do some special logic to make sure they have opportunity to be handle first
```
madolson, line 20: I think it would be useful to enumerate these here. Maybe everything that is not a data command? Reminds me a bit of this check: https://github.com/valkey-io/valkey/blob/5166d489da1bbf49643dd6b44cd2593e21c09dc3/src/server.c#L4061.

ADMIN-PORT.md
```diff
@@ -0,0 +1,44 @@
...
+2. Run some special commands through the connected IP address: Before we run every commands, we need check if the client ip address is in a linkedlist, it costs
+                                                               CPU and extra memory, it causes side effect in performance
+
+3. Iptables: It works on transport layer and IP address, but it can not apply to the specific commands of the Valkey
+
+Based on all above 3 candidate defects, we decide to select "admin-port" to implement our goal.
+
+
+## Specification 
+
```
madolson, line 37: I'm trying to think if there is anything else we're missing here. Do modules need to be able to differentiate this special port at ?

madolson (commented):
I'm generally aligned with this.

PingXie:
Thinking about this some more, I am not sure why ACL can't solve this problem. One could have a default user without 
a user name, which is granted a subset of commands; and then "admin" users are always required to enter their user names + password. We have been using this model just fine and it is also more secure, out of the box, IMO. The admin port would not be secure by default and the user will need to make sure their firewall is properly set up, which means more things to track.

There is one scenario though that I think an admin port could help. This is when the client application uses up `maxclients` such that my sidecar agent couldn't connect to the server to either gather metrics or perform management operations. So having a separate port could allow us to enforce a different `maxclients` policy but this new port doesn't need to have special power, hence less of a security concern.

ADMIN-PORT.md
hwware: yes, we could

ADMIN-PORT.md
hwware: Sorry @madolson this is an enhancement feature based on the admin-port added. The main feature is that the connection through admin-port could call some special command. If we have admin-port, we could have more fancy features later, including this.

ADMIN-PORT.md
hwware: I think we do not need to differentiate the module. Because the goal of importing this admin-port is to run some special commands or load some specific modules. We do not want to make the module as another outlier for the commands

ADMIN-PORT.md
```diff
@@ -0,0 +1,45 @@
...
+## Abstract
+
+In Valkey.conf, Add a new config parameter: "admin-port" to allow internal administrators in a company or special users from public internet to connect to the Valkey server to execute some special commands or load specfic modules.
+
+
+## Motivation
+
+For adminstrators, they may want to execute some special commands in the server, and other general clients are not allowed to run these kinds of commands.
+We want to let adminstrator connect to this admin port to do some special things.
```
zuiderkwast, line 16: There is something here that I still don't understand.

Which *special commands* exactly are you thinking about? An administrator can start the node with a config file or write the config on the command line. The administrator has access to the machine with ssh or something like that, right?

Is it something else you want to do?

Do you want to start all nodes in the same way, and later configure them for each customer?

Do you want the customer to have full access to all commands and even configure ACL, but still not be able to access something that only you (cloud operator) can access using the admin port?

hwware: image

PingXie: >However sometimes, admin do not want to this (maybe due to the uncertain username),

Could you elaborate on why they would like to avoid using ACLs? 

My guess is that rotating secrets adds complexity and management overhead. Without regular rotation, the password or secret becomes vulnerable to compromise. In that case, security is weakened because a static secret is bound to leak over time.

If that's the case, I could be convinced to implement a dedicated port restricted to the loopback interface. However, I think it should still integrate with the ACL subsystem. Specifically, we may need to introduce a hardcoded `admin` user with default ACL settings that can be overridden via configuration. Is that in line with your thinking?

xbasel: Perhaps use a dedicated Unix socket instead of an actual port and leave it to the customer to decide how to access it (it can be forwarded to a TCP port).

madolson:
Open questions:
1. Can remote clients opt-in to being marked as an admin client? Do we need to have multiple configurations to specify a separate network card?
2. Will unix domain sockets be considered the same as a local host for determining if a client should be marked as admin?

hwware:
> Open questions:
> 
> 1. Can remote clients opt-in to being marked as an admin client? Do we need to have multiple configurations to specify a separate network card?

     The admin client role is only decided by the internal administrator, whatever where the client comes from (even it is opt-in client).
     And we do not need the multiply configurations to specify a separate network card. We only need to care about the connection to the admin port.

> 2. Will unix domain sockets be considered the same as a local host for determining if a client should be marked as admin?
       Because the unix domain socket is a different protocol from tcp/ip, In this RFC, I do not want to  mix these 2 protocols into one way to check if this is an admin connection.  If necessary, we can consider unix domain socket to be another admin client in a separated feature. 
       
> 3. Should connecting to the admin endpoint automatically authenticate the client as an ACL user or should it just be a special flag.

       No, ACL and special flag are not considered for admin port. We only care the connection through the admin-port.

zuiderkwast:
I think this:

> 3\. Should connecting to the admin endpoint automatically authenticate the client as an ACL user or should it just be a special flag.
>         No, ACL and special flag are not considered for admin port. We only care the connection through the admin-port.

contradicts this:

> Administrator wants to execute some special commands in the servers, and general clients are not allowed to run these kinds of commands.

Why? If the client connected over the admin port doesn't have any more permissions than a  general client connected over the regular port, then it can't execute any more commands than the general client connected.

hwware:
> I think this:
> 
> > 3. Should connecting to the admin endpoint automatically authenticate the client as an ACL user or should it just be a special flag.
> > No, ACL and special flag are not considered for admin port. We only care the connection through the admin-port.
> 
> contradicts this:
> 
> > Administrator wants to execute some special commands in the servers, and general clients are not allowed to run these kinds of commands.
> 
> Why? If the client connected over the admin port doesn't have any more permissions than a general client connected over the regular port, then it can't execute any more commands than the general client connected.


First all, the initial goal of using admin-port is only exposed to internal administrator or cloud provider.  We strongly discourage to 
let general clients know it or use it.
Secondly,  in programming, the developer should first check if the connection is through admin-port, then check the ACL if it is available. If it is an admin-port connection, the special command can be executed definitely.

zuiderkwast:
> If it is an admin-port connection, the special command can be executed definitely.

There is nothing for this in this RFC and in the implementation PR.

If the config says `requirepass`, then the admin port client can't do anything without a password. What I can find in this RFC and in the implementation, it has no special permissions at all.

soloestoy:
In my understanding, there are two main issues:

1. Administrators must be able to access valkey under any circumstances, meaning they should not be limited by the `maxclients` limit (currently, we only limit the number of connections, not CPU or bandwidth).

    Using an admin port is a reasonable solution.

2. Administrators need high-level permissions to manage valkey.

    I believe ACL is more suitable for this.

ADMIN-PORT.md
```diff
@@ -0,0 +1,46 @@
...
+3. Iptables: It works on transport layer and IP address, but it can not apply to the specific commands of the Valkey
+
+Based on all above 3 candidate defects, we decide to select "admin-port" to implement our goal.
+
+
+## Specification
+
+1. The admin client role is only decided by the cloud provides or internal administrators, whatever where the client comes from. Thus generally, admin-port should not be exposed to public.
+2. If a client connects via admin-port and ACL rules apply to this client as well, this client should follow the ACL to execute some commands
+3. If a client connects via non-admin-port and ACL rules apply to this client as well, this client is not allowed to execute those special commands.
```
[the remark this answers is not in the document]
hwware: Update the doc, and list 2 special command examples there, pls take a look.  Thanks your comments

ADMIN-PORT.md
```diff
@@ -0,0 +1,46 @@
...
+
+
+## Motivation
+
+For adminstrators, they may want to execute some special commands in the server, which are usually defined in the module, and
+other general clients are not allowed to run these kinds of commands. Thus connecting to an admin-port is a good way to implement this feature. 
+
+Beside above mentioned, we could implement the following features in the future once admin-port is commited:
+1. The connections through admin port are not limited to maxclient number, we can always guarantee the administrators can connect to the server.
+2. Some status check commands could be sent and receied through admin port, we can do some special logic to make sure they have opportunity to be handle first, such as: PING, HELLO, ECHO, INFO etc
```
zuiderkwast, line 20: Are you suggesting we add some code like this?

```C
if (c->cmd->proc == pingCommand || 
    c->cmd->proc == helloCommand ||
    c->cmd->proc == echoCommand ||
    c->cmd->proc == infoCommand) {
    /* Permission OK */
}
```

I don't like a hard-coded list of commands like this. It is better that it matched to one ACL category or something like that. Or maybe configurable.

zuiderkwast: Then I think the RFC should describe how the module can check this. We need a new module API? Or a command flag to ValkeyModule_CreateCommand?

hwware: In the Specification part, I add several words: You can find a way to to get the port number that client connects to the server and then to decide if the command can be called by the client.

In fact, developers have a lot of ways to get the client connected port, and in our side, we have one way to check the client connected node ipaddress and port. And I believe there are other ways to implement the same function. Thus, in this RFC, I just want to indicate the port check should happen in the function which is defined in RedisModule_CreateCommand

zuiderkwast: I don't know how a module get this information. I think it's good to describe some example how it can be done in the design considerations. It is useful context to understand the feature.

You have to call `ValkeyModule_Call("CLIENT INFO")`?

I would hope the module can find out with `ValkeyModule_GetClientInfoById` but it seems to return only the peer port and IP, not the server port it connected to.

<details>
<summary>ValkeyModule_GetClientInfoById</summary>

```
              int ValkeyModule_GetClientInfoById(void *ci, uint64_t id);

       Available since: 6.0.0

       Return information about the client with the  specified  ID  (that  was
       previously  obtained  via  the ValkeyModule_GetClientId() API).  If the
       client exists, VALKEYMODULE_OK is returned, otherwise  VALKEYMODULE_ERR
       is returned.

       When  the  client exist and the ci pointer is not NULL, but points to a
       structure of type ValkeyModuleClientInfoV1, previously initialized with
       the correct VALKEYMODULE_CLIENTINFO_INITIALIZER_V1,  the  structure  is
       populated with the following fields:

               uint64_t flags;         // VALKEYMODULE_CLIENTINFO_FLAG_*
               uint64_t id;            // Client ID
               char addr[46];          // IPv4 or IPv6 address.
               uint16_t port;          // TCP port.
               uint16_t db;            // Selected DB.

       Note:  the  client  ID is useless in the context of this call, since we
       already know, however the same structure could be used  in  other  con‐
       texts  where we don’t know the client ID, yet the same structure is re‐
       turned.

       With flags having the following meaning:

              VALKEYMODULE_CLIENTINFO_FLAG_SSL          Client using SSL connection.
              VALKEYMODULE_CLIENTINFO_FLAG_PUBSUB       Client in Pub/Sub mode.
              VALKEYMODULE_CLIENTINFO_FLAG_BLOCKED      Client blocked in command.
              VALKEYMODULE_CLIENTINFO_FLAG_TRACKING     Client with keys tracking on.
              VALKEYMODULE_CLIENTINFO_FLAG_UNIXSOCKET   Client using unix domain socket.
              VALKEYMODULE_CLIENTINFO_FLAG_MULTI        Client in MULTI state.

       However passing NULL is a way to just check if  the  client  exists  in
       case we are not interested in any additional information.

       This  is  the  correct usage when we want the client info structure re‐
       turned:

               ValkeyModuleClientInfo ci = VALKEYMODULE_CLIENTINFO_INITIALIZER;
               int retval = ValkeyModule_GetClientInfoById(&ci,client_id);
               if (retval == VALKEYMODULE_OK) {
                   printf("Address: %s\n", ci.addr);
               }
```
</detail>

hwware: I just add a block: Sample for Module, which indicates one way to get the admin port and how to use it in module. But I think more way to get it. Please take a look

ADMIN-PORT.md
```diff
@@ -0,0 +1,87 @@
...
+        int getClientConnectedPort(client *c) {
+            .........
+            return c->conn->connected_port;
+        } 
+
+step 3: create an API in module:
+        int VM_GetClientConnectedPort(ValkeyModuleCtx *ctx) {
+            return getClientConnectedPort(ctx->client);
+        }
+        Thus, module developers can call it to get the client connected server port.
```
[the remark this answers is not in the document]
zuiderkwast: Is it an example or is it part of the specification?

If this module API (to detect if the client is using the admin port or not) is only an example, a future idea, it means that in this RFC there is not yet any way to detect if a client is using the admin port.

I don't like the feature if it is only useful when it is combined with some future feature that is not yet discussed. We should include the module API in the RFC too, or some other way to grant a client some permissions based on admin port.

hwware:
Because I decide to narrow down the scope of admin port parameter in Valkey, I decide to close this RFC and describe the details in the PR https://github.com/valkey-io/valkey/pull/1120

PR #4, opened 2024-07-25
BloomFilter Module support for Valkey RFC

KarthikSubbarao:
Valkey-Bloom (BSD-3-Clause) is a Rust Valkey-Module which brings a native and space efficient probabilistic Module data type to Valkey. With this, users can create filters (space-efficient probabilistic Module data type) to add elements, perform “check” operation to test whether an element exists, auto scale their filters, perform RDB Save and load operations, etc.

It is compatible with the BloomFilter (BF.*) command APIs of the ReBloom Module from Redis Ltd.


Bloom.md
```diff
@@ -0,0 +1,167 @@
...
+
+The following Bloom Filter commands are supported and their API syntax is compatible with ReBloom:
+BF.ADD <key> <item>
+BF.EXISTS <key> <item>
+BF.MADD <key> <item> [<item> ...]
+BF.MEXISTS <key> <item> [<item> ...]
+BF.CARD <key>
+BF.INFO <key> [CAPACITY | SIZE | FILTERS | ITEMS | EXPANSION]
+BF.RESERVE <key> <false_positive_rate> <capacity> [EXPANSION <expansion>] | [NONSCALING]
+BF.INSERT <key> [ERROR <fp_error>] [CAPACITY <capacity>] [EXPANSION <expansion>] [NOCREATE] [NONSCALING] ITEMS <item> [<item> ...]
```
hwware, line 126: Can you double check if the ValkeyBloom has same command syntax with Redis Bloom，because ReBloom syntax is: 
 BF.INSERT key [CAPACITY capacity] [ERROR error]
  [EXPANSION expansion] [NOCREATE] [NONSCALING] ITEMS item [item
  ...]  https://redis.io/docs/latest/commands/bf.insert/

KarthikSubbarao: ReBloom's BF.INSERT command accepts any order of these optional arguments.

ValkeyBloom also does the same.

I can update the documentation so that it does not cause any confusion regarding this

Bloom.md
```diff
@@ -0,0 +1,199 @@
...
+
+Redis Ltd.‘s ReBloom is published under a proprietery license and hence cannot be distributed freely with ValKey.
+
+There is growing [demand](https://github.com/orgs/valkey-io/discussions?discussions_q=+bloom+) for an (1) Open Source bloom filter feature in Valkey which is (2) compatible with the ReBloom API syntax and with existing ReBloom based client libraries. ValkeyBloom can help address both these requirements.
+
+
+## Design Considerations
+
+The ValkeyBloom module brings in a bloom module data type into Valkey and provides commands to create / reserve bloom filters, operate on them (add items, check if items exist), inspect bloom filters, etc. It allows customization of properties of bloom filter (capacity, false positive rate, expansion rate, specification of scaling vs non-scaling filters etc) through commands and configurations. It also allows users to create scalable bloom filters and back up & restore bloom filters (through RDB load and save).
```
zuiderkwast, line 26: This is a description of what's already implemented and how it behaves. I suppose it's meant as a spec for what we want, so I think most of it belongs under Specification rather than Design considerations.

What we want under design consiserations is basically what we have to chose between in the case of different possibilities (like for aof rewrite).

Bloom.md
```diff
@@ -0,0 +1,199 @@
...
+
+The data above is written into the RDB during save operations and can be restored from RDB load. However, the data that gets written to the RDB is specific to the data type's structure and struct members. Additionally, the data within the underlying bloom filter (from the external crate) is specific to the implementation of the bloom filter as the hash key (seed), raw bit array data, etc. can all vary.
+
+Because of this, it is not possible to be RDB compatible with ReBloom. It might be possible if the RDB Save/Load of the ReBloom module is reverse engineered AND also the bloom filter implementation is changed to follow ReBloom Module's raw Bloom Filter structure syntax. Additionally, if there is any RDB change in ReBloom, this reverse engineering + structure updates will be needed once again.
+
+For these reasons, ValkeyBloom is not RDB Compatible with ReBloom.
+
+#### AOF Rewrite handling:
+
+Module data types (including bloom) can implement a callback function that will be triggered for Bloom objects to rewrites its data as command/s. Currently, we have not yet implemented the AOF callback and we can discuss AOF handling options here.
```
zuiderkwast, line 71: I suppose it's fine to add the proposal with open queations like this, but IMO it can't be Accpeted before we have made a deciion about it and edited this text.

zuiderkwast (commented):
Looks good, maybe a bit longer than necessary but it's good that all details are included. I think some content should be moved to a different section though.

I'm missing a discussion on whether or not we need RDB format compatibility (DUMP/RESTORE) with ReBloom. Personally, I don't think we need it, but I think it needs to be mentioned and something about why. (Possibility to support data migration from rebloom?)

Regarding AOF, it can be mentioned that not everyone uses RDB and what happens in such setups.

KarthikSubbarao:
> I'm missing a discussion on whether or not we need RDB format compatibility (DUMP/RESTORE) with ReBloom. Personally, I don't think we need it, but I think it needs to be mentioned and something about why. (Possibility to support data migration from rebloom?)

I have documented this under the "RDB Compatibility with ReBloom" section

Bloom.md
```diff
@@ -0,0 +1,240 @@
...
+
+1. We can handle AOF rewrite by saving commands such as BF.RESERVE. With this, we will be able to re-create a Bloom Object with the same properties (expansion, capacity, false positive rate). However, the bits will not be set and when restored from the AOF, no items would "exist" (be set) on the bloom object. This can be considered data loss.
+
+This is not an issue with RDB Load and Save because we save the Bloom object's raw Bloom filter bit array data during RDB Save and we are able to restore this during RDB Load.
+
+2. For AOF rewrite to support saving the exact state of the Bloom object (including the items that were "set"), we need to include the dump in the AOF and will need to support a command that can restore this data. ReBloom supports a BF.LOADCHUNK command to restore a bloom object in interations from its dump.
+
+### Migrating workloads from ReBloom:
+
+Customers that currently use ReBloom can move to ValkeyBloom using two approaches:
```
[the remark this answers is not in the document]
KarthikSubbarao: Sure. I will add these details

Bloom.md
KarthikSubbarao: @zuiderkwast - This section has been updated and I documented an approach that we plan to implement for AOF Rewrite

Bloom.md
```diff
@@ -0,0 +1,248 @@
+---
+RFC: 4
+Status: Proposed
+---
+
+# ValKeyBloom Module RFC
```
madolson, line 6:
```suggestion
# ValkeyBloom Module RFC
```
K is not capitalized in Valkey.

Bloom.md
```diff
@@ -0,0 +1,248 @@
...
+
+## Abstract
+
+The proposed feature is ValKeyBloom which is a Rust based Module that brings a native bloom filter data type into Valkey.
+The module should be API-compatible with Redis Ltd.’s ReBloom Module.
+
+## Motivation
+
+Bloom filters are a space efficient probabilistic data structure that can be used to “check” whether an element exists in a set (with a defined false positive), and to “add” elements to a set. While checking whether an item exists, false positives are possible, but false negatives are not possible. https://en.wikipedia.org/wiki/Bloom_filter
```
madolson, line 15:
```suggestion
Bloom filters are a space efficient probabilistic data structure that can be used to “check” whether an element exists in a set (with a defined false positive), and to “add” elements to a set.
While checking whether an item exists, false positives are possible, but false negatives are not possible. https://en.wikipedia.org/wiki/Bloom_filter
```
It makes it easier to read if paragraph lines are broken up. Either reflow at some line length (like 120 characters) or break on each period.

Bloom.md
```diff
@@ -0,0 +1,248 @@
...
+Currently following commands (from ReBloom) are not supported:
+BF.LOADCHUNK <key> <iterator> <data>
+BF.SCANDUMP <key> <iterator>
+
+The BF.SCANDUMP command is used to perform an incremental save on specific Bloom filter object.
+The BF.LOADCHUNK is used to incrementally load / restore a bloom filter object from data from the BF.SCANDUMP command.
+
+The reason for not implementing these two commands is because the Module provides the ability to load and save BloomModule data type items during RDB load and save. BF.LOADCHUNK and BF.SCANDUMP are APIs to load bloom filter objects through commands, but since we will provide RDB save & load, having specific commands for the same purpose was not considered as required.
+
+Note: Based on the AOF Rewrite discussion, we will decide on whether these commands should be supported.
```
madolson, line 218: Does scandump work for AOF? We need to dump the entire object as a single object int he AOF file. Could we instead just have a BF.LOAD command? I suppose we should figure out what RedisBloom does.

Bloom.md
```diff
@@ -0,0 +1,248 @@
...
+
+The following module configurations are added to customize bloom filters:
+1. bf.bloom_capacity: Controls the default capacity. When create operations (BF.ADD/MADD) are used, bloom objects created will use the capacity specified by this config. This controls the number of items that can be added to a bloom filter object before it is scaled out (scaling) OR before it rejects add operations due to insufficient capacity on the object (nonscaling).
+2. bf.bloom_expansion_rate: Controls the default expansion rate. When create operations (BF.ADD/MADD) are used, bloom objects created will use the expansion rate specified by this config. This controls the capacity of the new filter that gets added to the list of filters as a result of scaling.
+3. bf.bloom_max_filters: Controls the maximum number of filters that any bloom object is allowed to scale out to.
+4. bf.bloom_fp_rate: Controls the default false positive rate that new bloom objects created (from BF.ADD/MADD) will use.
+5. bf.bloom_large_item_threshold: Memory usage of a bloom object beyond which bloom objects are exempted from defrag and are forced to be async freed (when deleted).
```
madolson, line 229:
```suggestion
1. bf.bloom-capacity: Controls the default capacity. When create operations (BF.ADD/MADD) are used, bloom objects created will use the capacity specified by this config. This controls the number of items that can be added to a bloom filter object before it is scaled out (scaling) OR before it rejects add operations due to insufficient capacity on the object (nonscaling).
2. bf.bloom-expansion-rate: Controls the default expansion rate. When create operations (BF.ADD/MADD) are used, bloom objects created will use the expansion rate specified by this config. This controls the capacity of the new filter that gets added to the list of filters as a result of scaling.
3. bf.bloom-max-filters: Controls the maximum number of filters that any bloom object is allowed to scale out to.
4. bf.bloom-fp-rate: Controls the default false positive rate that new bloom objects created (from BF.ADD/MADD) will use.
5. bf.bloom-large-item-threshold: Memory usage of a bloom object beyond which bloom objects are exempted from defrag and are forced to be async freed (when deleted).
```

Are these based of RedisBloom configs, or are these are own configs?
Why is 5 a configuration? That seems like something we should control.

Bloom.md
```diff
@@ -0,0 +1,248 @@
...
+    valkey-cli config set notify-keyspace-events KEA
+2. suscribe to keyspace & keyevent event channels:
+    valkey-cli psubscribe '__key*__:*'
+```
+
+### Scalable filters / Non Scaling filters
+
+Bloom Filters can either be configured as scalable (default) or non-scalable through specification with the BF.RESERVE or BF.INSERT command.
+
+When non-scaling filters reach their capacity, if a user tries to add items to the bloom object, an error is returned.
```
madolson, line 129: Why is an error returned? Wouldn't it also be valid to just return an OK back? If someone is aware of the tradeoffs associated with a bloom filter, that is can return false negatives, it seems OK that we could saturate it.

Bloom.md
```diff
@@ -0,0 +1,248 @@
...
+* free: Frees the bloom filter object when it is deleted, expired or evicted.
+* defrag: Supports active defrag for Bloom filter objects
+* mem_usage: Reports bytes used by a Bloom object and is used by the MEMORY USAGE command
+* copy: Supports a deep copy of bloom filter objects and is used by the COPY command
+* free_effort: Determine whether the bloom object's memory needs to be lazy reclaimed or synchronously freed. We return the number of filters in the bloom object as the free effort and this is similar to how the core handles free_effort for aggregated objects.
+
+
+### Replication
+
+Every Bloom Filter based write command will be replicated to replica nodes.
```
madolson, line 108: If a write does not mutate the object, for example the same item has the same fingerprint, do we still replicate the command?

KarthikSubbarao: No, the highlighted line should be worded to clarify this. I will update it. Only write operations (bloom object creations, scaling, setting of an item on a filter which returns 1 indicating a new entry, etc) will be replicated. Attempts of adding an already existing item to a bloom object will not replicated

Bloom.md
KarthikSubbarao: > Why is an error returned?

This is the behavior built into ReBloom based on which this error logic was implemented. 

> Wouldn't it also be valid to just return an OK back? If someone is aware of the tradeoffs associated with a bloom filter, that is can return false negatives, it seems OK that we could saturate it.

Originally, I felt the same. But this can lead to the user observing high error_rate. It might be better for the user to be notified of this happening.

If we think that saturated bloom filters (with low occurrences of "new items" being added) are a use case, we can consider this being a configurable behavior

Bloom.md
KarthikSubbarao: > Are these based of RedisBloom configs, or are these are own configs?

These are five configurations that are of own.

> Why is 5 a configuration? That seems like something we should control.

Do you mean we should have a static constant defined based on which we can handle large items differently (defrag, etc)? (Instead of making this configurable)

Bloom.md
madolson: Let's add that as part of the spec then. I don't necessarily want to copy RedisBloom if they made bad decisions.

Bloom.md
KarthikSubbarao: Sorry, this was a typo. I fixed it.

Bloom.md
madolson: > Do you mean we should have a static constant defined based on which we can handle large items differently (defrag, etc)? (Instead of making this configurable)

Yeah. I'm not familiar with other places where this is tunable.

Bloom.md
```diff
@@ -0,0 +1,248 @@
...
+Certain libraries (e.g. fastbloom) are no longer compatible with older versions (after updates) resuling in serialization / deserialization incompatibility.
+
+growable-bloom-filter is a bloom filter library that implements a scalable bloom filter. However, it does not allow control over when the scaling occurs and does not expose APIs to check the number of filters that it has scaled out to. This crate also has had a breaking change across versions and bloom filters of the older version are not loadable on the newer versions.
+
+
+### Large BloomFilter objects
+
+Create and Delete operations on large bloom filters take longer durations and will block the main thread for this duration. 
+Because of this, the following operations can be handled differently:
+* defrag callback: If the memory used is greater than X MB (configurable with `bf.bloom_large_item_threshold`), we can skip defrag operations on this bloom object.
```
madolson, line 190: Can we not iteratively defrag large bloom objects?

Bloom.md
```diff
@@ -0,0 +1,248 @@
...
+* free_effort callback: This callback decides the free effort for the bloom object. If it is greater than X MB (configurable with `bf.bloom_large_item_threshold`), we can return 0 to use async free logic.
+* create operations (BF.ADD/MADD/INSERT): We can compute the estimated size of the bloom filter that will be created as a result of the operation. We can consider rejecting requests in case of unavailable memory.
```
madolson, line 192: How long can BF.INSERT really take?

Bloom.md
```diff
@@ -0,0 +1,248 @@
...
+
+1. Create the bloom filter objects to have the same properties using BF.RESERVE / BF.INSERT on Valkey (with ValkeyBloom loaded). Re-populate the bloom filter objects by inserting items by moving the existing bloom workload to the Valkey server (with ValkeyBloom loaded). The workload can be moved without any errors since we are API Compatible.
+
+Pros
+* This is the simplest option in terms of effort - assuming the user is fine with recreating bloom objects and adding items on the Valkey server (with ValkeyBloom).
+
+Cons
+* The user will need to re-create the bloom objects (using BF.RESERVE/BF.INSERT) and populate these objects **AFTER** switching to Valkey (with ValkeyBloom).
+
+2. Users can generate an AOF file from a server (that has the ReBloom module loaded) and with an on-going bloom workload that creates bloom filters & inserts items into them. Next, this can be re-played on a Valkey Server (with ValkeyBloom loaded). Then, the user can move their existing bloom workload to the Valkey server (with ValkeyBloom loaded).
```
madolson, line 87: AOF will eventually need to be truncated, I don't think this is really a tenable approach for migration. Bloom filters are intended to be lossy.

Maybe we should instead consider a way to load a bloom module with its hashes, but use our own hash algorithm when we are generating our own? We might need new types of module APis for that though.

Bloom.md
```diff
@@ -0,0 +1,248 @@
...
+
+### Replication
+
+Every Bloom Filter based write command will be replicated to replica nodes.
+
+### Keyspace Event Notification
+
+Every bloom filter based write command will be made to publish a keyspace event after the data is mutated.
+* Event type: VALKEYMODULE_NOTIFY_GENERIC
+* Event name: command name in lowercase. e.g., BF.ADD command publishes event "bloom.add".
```
madolson, line 114: Redis historically didn't generate the command name in lowercase, but generated a simplified version. For example, bf insert might get reduced to a bf.reserve + bf.add so that clients only need to listen for bf.add and bf.reserve and can ignore bf.insert.

Bloom.md
KarthikSubbarao: That makes sense. I added line length based new line breaks at 120 characters

Bloom.md
KarthikSubbarao: Since we are using an external crate for the Bloom Library, we do not have direct access to the bit vector and hash functions. If we did, we could iteratively defragment this to avoid one big re-allocation.

We could optimize on scalable bloom filters to support iteratively defrag (delete and re-create) each bloom filter in the vector of boom filters. I think this is the most we can do

Bloom.md
KarthikSubbarao: Sounds good. This can be changed to a static value. I will list some of the statics like this in a section below

Bloom.md
KarthikSubbarao: Yes, Scalable Bloom filters will grow in used memory after creation - but only as a result of a BF.ADD, BF.MADD, or BF.INSERT operation. We can consider rejecting these requests if we identify that it will result in a scale out operation OR beyond a certain number of filters

Bloom.md
KarthikSubbarao: Sounds good. I can make a condensed list of events for the various APIs we support

Bloom.md
KarthikSubbarao: >AOF will eventually need to be truncated, I don't think this is really a tenable approach for migration.

I agree, this will not work when there are truncations in AOF or rewrites.

>Bloom filters are intended to be lossy.

This is true. Customers also use bloom filters with TTLs and these customers should be ok with recreating bloom objects after migrating to Valkey

Bloom.md
```diff
@@ -0,0 +1,341 @@
...
+# ValkeyBloom Module RFC
+
+## Abstract
+
+The proposed feature is ValkeyBloom which is a Rust based Module that brings a native bloom filter data type into Valkey.
+The module should be API-compatible with Redis Ltd.’s ReBloom Module.
+
+## Motivation
+
+Bloom filters are a space efficient probabilistic data structure that can be used to “check” whether an element exists in
```
madolson, line 15:
```suggestion
Bloom filters are a space efficient probabilistic data structure that can be used to "check" whether an element exists in
```
I just realized these are curly quotes. Probably fine to use them, but it would make sense to keep this all as ascii.

Bloom.md
```diff
@@ -0,0 +1,341 @@
...
+RFC: 4
+Status: Proposed
+---
+
+# ValkeyBloom Module RFC
+
+## Abstract
+
+The proposed feature is ValkeyBloom which is a Rust based Module that brings a native bloom filter data type into Valkey.
+The module should be API-compatible with Redis Ltd.’s ReBloom Module.
```
bbarani, line 11: "The module should be API-compatible with **current** version of  Redis Ltd.’s ReBloom" Module?

Bloom.md
```diff
@@ -0,0 +1,341 @@
...
+
+The data above is written into the RDB during save operations and can be restored from RDB load. However, the data that
+gets written to the RDB is specific to the data type's structure and struct members. Additionally, the data within the
+underlying bloom filter (from the external crate) is specific to the implementation of the bloom filter as the
+hash key (seed), hashing algorithm, raw bit array data, etc. can all vary.
+
+Because of this, it is not possible to be RDB compatible with ReBloom. It might be possible if the RDB Save/Load of the
+ReBloom module is reverse engineered AND also the bloom filter implementation is changed to follow ReBloom Module's raw
+Bloom Filter structure syntax. Additionally, if there is any RDB change in ReBloom, this reverse engineering + structure
+updates will be needed once again.
```
hpatro, line 97: I think it's going to be a pain maintaining RDB compatibility across version with Redis. I feel the migration should be handled externally if possible (convert RDB file to commands and send to valkey).

Bloom.md
```diff
@@ -0,0 +1,341 @@
...
+
+
+Users can subscribe to the bloom events via the standard keyspace event pub/sub. For example,
+
+```text
+1. enable keyspace event notifications:
+    valkey-cli config set notify-keyspace-events KEA
+2. suscribe to keyspace & keyevent event channels:
+    valkey-cli psubscribe '__key*__:*'
+```
```
hpatro, line 179: This is a generic mechanism for all keys. Do you need to implement anything specifically for Bloom module?

Bloom.md
```diff
@@ -0,0 +1,341 @@
...
+### RDB Compatibility with ReBloom:
+
+During RDB Save of a bloom object, we will need to save the number of filters, expansion rate, false positive rate.
+And for every underlying bloom filter in this object, we store its hash keys used by hashing functions, number of
+hashing functions, number of bits of the bit array, bytes of the bit array itself.
+
+The data above is written into the RDB during save operations and can be restored from RDB load. However, the data that
+gets written to the RDB is specific to the data type's structure and struct members. Additionally, the data within the
+underlying bloom filter (from the external crate) is specific to the implementation of the bloom filter as the
+hash key (seed), hashing algorithm, raw bit array data, etc. can all vary.
```
madolson, line 92: Did we decide if we are randomly generating a seed or not?

Bloom.md
```diff
@@ -0,0 +1,341 @@
...
+
+The following Bloom Filter commands are supported and their API syntax is compatible with ReBloom:
+BF.ADD <key> <item>
+BF.EXISTS <key> <item>
+BF.MADD <key> <item> [<item> ...]
+BF.MEXISTS <key> <item> [<item> ...]
+BF.CARD <key>
+BF.INFO <key> [CAPACITY | SIZE | FILTERS | ITEMS | EXPANSION]
+BF.RESERVE <key> <false_positive_rate> <capacity> [EXPANSION <expansion>] | [NONSCALING]
+BF.INSERT <key> [CAPACITY <capacity>] [ERROR <fp_error>] [EXPANSION <expansion>] [NOCREATE] [NONSCALING] ITEMS <item> [<item> ...]
```
hpatro, line 291: Would prefer request / response for each command in the specification.

Bloom.md
```diff
@@ -0,0 +1,341 @@
...
+4. bf.bloom_fp_rate: Controls the default false positive rate that new bloom objects created (from BF.ADD/MADD) will use.
+
+### Constants
+1. bloom_large_item_threshold: Memory usage of a bloom object beyond which bloom objects are exempted from defrag operations
+    and when deleted, the Module will indicate the object's free_effort as 0 to be async freed.
+
+
+### ACL
+
+The ValkeyBloom module will introduce a new ACL category - @bloom.
```
hpatro, line 327: Is this aligned with rebloom ?

Bloom.md
```diff
@@ -0,0 +1,341 @@
...
+
+There are 4 existing ACL categories which are updated to include new BloomFilter commands: @read, @write, @fast, @slow.
+
+### Info metrics
+
+Info metrics are visible through the “info bloom or “info modules” command:
+* Number of bloom filter objects
+* Total bytes used by bloom filter objects
```
hpatro, line 335: Are there any metrics which would be handy for administrator / debugging ?

Bloom.md
```diff
@@ -0,0 +1,341 @@
...
+When scaling filters reach their capacity, if a user adds an item to the bloom object, a new bloom filter is created and
+added to the list of bloom filters within the same bloom object. This new bloom filter will have a larger capacity
+(previous bloom filter's capacity * expansion rate of the bloom object).
+
+When we want to check whether an item exists on a bloom object (BF.EXISTS/BF.MEXISTS), we look through each filter
+(from oldest to newest) in the object's filter list and perform a check operation on each one. Similarly, to add a new
+item to the bloom object, we check through all the filters to see if the item already exists and if not, the item is
+added to the current filter.
+
+When a bloom object has a larger number of bloom filters, it will result in reduced performance.
```
bbarani, line 204: Do we have any benchmarking metrics?

Bloom.md
```diff
@@ -0,0 +1,341 @@
...
+
+The reason for not implementing these two commands is because the Module provides the ability to load and save BloomModule
+data type items during RDB load and save. BF.LOADCHUNK and BF.SCANDUMP are APIs to load bloom filter objects through commands,
+but since we will provide RDB save & load, having specific commands for the same purpose was not considered as required.
+
+Note: Based on the AOF Rewrite discussion, we will decide on whether these commands should be supported.
+
+### Configurations
+
+The default properties using which Bloom Filter objects are created can be controlled using configs.
```
hpatro, line 308: Could we also mention that these can be overridden for each bloom object via BF.RESERVE / BF.INSERT

Bloom.md
```diff
@@ -0,0 +1,341 @@
...
+* Hash Key: Keys used by the hashing function.
+
+
+### Module OnLoad
+
+Upon loading, the module registers a new bloom module based data type, creates bloom filter (BF.*) commands,
+bloom specific configurations and the bloom ACL category.
+
+* Module name: bloom
+* Data type name: bloom
```
hpatro, line 73: Do we handle module unload ?

Bloom.md
```diff
@@ -0,0 +1,341 @@
...
+* Hash Functions: Hash functions used for bit check and set operations underneath.
+* Hash Key: Keys used by the hashing function.
+
+
+### Module OnLoad
+
+Upon loading, the module registers a new bloom module based data type, creates bloom filter (BF.*) commands,
+bloom specific configurations and the bloom ACL category.
+
+* Module name: bloom
```
hpatro, line 72: Can this conflict with other oss bloom module ? would prefer `valkeybloom`

Bloom.md
```diff
@@ -0,0 +1,341 @@
...
+* free_effort: Determine whether the bloom object's memory needs to be lazy reclaimed or synchronously freed. We return
+    the number of filters in the bloom object as the free effort and this is similar to how the core handles free_effort
+    for aggregated objects.
+
+
+### Replication
+
+Every Bloom Filter based write operation (bloom object creations, scaling, setting of an item on a filter which returns 1
+indicating a new entry, etc) will be replicated to replica nodes. Attempts of adding an already existing item to a bloom
+object (which returns 0) will not replicated.
```
hpatro, line 159: The configuration mentioned below can vary across primary/replica. How will it affect the replication stream then?

hpatro: For e.g., if the config value for `bf.bloom_fp_rate`  is set differently on primary and replica. Operation like `BF.ADD` might succeed on primary and fail on replica and vice versa.

hpatro: On this point, I think we should remove external facing configuration and set default value in the code and let it be overridable for each bloom filter key. WDYT ?

Bloom.md
```diff
@@ -0,0 +1,341 @@
...
+scaling vs non-scaling filters etc) through commands and configurations. It also allows users to create scalable bloom
+filters and back up & restore bloom filters (through RDB load and save).
+
+ValkeyBloom provides commands (BF.*), configs, etc to operate on Bloom objects which are top-level structures containing
+lower-level BloomFilter provided by an external crate. ValkeyBloom utilizes an open source (BSD-2) Bloom Filter Rust library
+around which it implements a scalable bloom filter.
+
+When a bloom filter is created, a bit array is created with a length proportional to the capacity (number of items the user
+wants to add to the filter) and hash functions are also created. The number of hash functions are controlled by the
+false positive rate that the user configures.
```
parthpatel, line 44: how many hash functions will a filter generally have? What does this function look like? ((ax+b)mod p)mod m) or something more complex?

Bloom.md
```diff
@@ -0,0 +1,341 @@
...
+
+When a user adds an item (e.g. BF.ADD) to the filter, the item is passed through the filter and corresponding bit are set
```
parthpatel, line 46: what is the filter? Is it just the hash function or something more?

Bloom.md
```diff
@@ -0,0 +1,341 @@
...
+to 1. When a user checks whether an item exists on a filter (e.g. BF.EXISTS), the item is passed through the filters
+and if all the resolved bits have values as 1, we can say that the item exists with a false positive rate of
+X (specified by the user when creating the filter). If any of the bits are 0, the item does not exist and the BF.EXISTS
```
parthpatel, line 49: Any external link explaining the calculation of this probability?

Bloom.md
```diff
@@ -0,0 +1,341 @@
...
+operation will return 0.
+
+We have the following terminologies / properties:
+* Bloom Object: The top level structure representing the data type. It contains meta data and a list of lower-level
+                Bloom Filters (Implemented by an external crate) in case of scaling and a single lower-level bloom filter
```
parthpatel, line 54: What does "lower-level bloom filters" mean? Do you just mean user initialized instance of bloom filter? It is not clear what is different during scaling vs non-scaling?

Bloom.md
```diff
@@ -0,0 +1,341 @@
...
+                in case of non scaling.
+* Bloom Filter: A single bloom filter (Implemented by an external crate).
+* Capacity: The number of items we expect to add to a bloom filter. This controls the size of the filter.
```
parthpatel, line 57: number of unique items? 
I think you meant to say, number of unique items that we want to be able to detect without false positives?

Bloom.md
```diff
@@ -0,0 +1,341 @@
...
+* False Positive Rate: The accuracy the user expects when operating (set / check) on a filter. This controls the number
```
parthpatel, line 58: If false positive rate is 0.1, how do I interpret it?

Bloom.md
```diff
@@ -0,0 +1,341 @@
...
+* Expansion Rate: This is used in scalable bloom filters where multiple bloom filters are stacked to allow users to
+                    continue using the same bloom object when it reaches capacity by adding another filter of larger
+                    capacity (expansion rate * prev filter capacity).
+* Hash Functions: Hash functions used for bit check and set operations underneath.
+* Hash Key: Keys used by the hashing function.
+
+
+### Module OnLoad
+
+Upon loading, the module registers a new bloom module based data type, creates bloom filter (BF.*) commands,
```
parthpatel, line 69: you mean "registers bloom filter..."?

Bloom.md
```diff
@@ -0,0 +1,341 @@
...
+Because of this, it is not possible to be RDB compatible with ReBloom. It might be possible if the RDB Save/Load of the
+ReBloom module is reverse engineered AND also the bloom filter implementation is changed to follow ReBloom Module's raw
+Bloom Filter structure syntax. Additionally, if there is any RDB change in ReBloom, this reverse engineering + structure
+updates will be needed once again.
+
+For these reasons, ValkeyBloom is not RDB Compatible with ReBloom.
+
+### AOF Rewrite handling:
+
+Module data types (including bloom) can implement a callback function that will be triggered for Bloom objects to rewrites
```
parthpatel, line 103: typo - rewrite

Bloom.md
```diff
@@ -0,0 +1,341 @@
...
+its data as command/s. From the AOF callback, we will handle AOF rewrite by saving a RESTORE command with the key, TTL, and
+serialized value of the corresponding bloom object.
```
parthpatel, line 105: how do we serialize the bitmap? what about hash functions?

Bloom.md
```diff
@@ -0,0 +1,341 @@
...
+
+There is an alternate to the approach above. It involves having the ValkeyBloom Module support BF.LOADCHUNK and
```
parthpatel, line 107: isn't the main goal for AOF rewrite to compress the consecutive commands on same key? Is the proposal to convert multiple consecutive bloom commands into a single restore command? An example would help understand this better.

Bloom.md
```diff
@@ -0,0 +1,341 @@
...
+* mem_usage: Reports bytes used by a Bloom object and is used by the MEMORY USAGE command
+* copy: Supports a deep copy of bloom filter objects and is used by the COPY command
+* free_effort: Determine whether the bloom object's memory needs to be lazy reclaimed or synchronously freed. We return
+    the number of filters in the bloom object as the free effort and this is similar to how the core handles free_effort
+    for aggregated objects.
+
+
+### Replication
+
+Every Bloom Filter based write operation (bloom object creations, scaling, setting of an item on a filter which returns 1
```
parthpatel, line 157: is it the actual operation with original data which replica has to re-hash, or just the location of the bits set as part of the operation?

Bloom.md
```diff
@@ -0,0 +1,341 @@
...
+| 70  | 2361183241434822608686                |
+| 75  | 75557862452714323477986               |
+| 100 | 2535301200456458804968143894986       |
+
+Practically, we can expect scaling to stop well before the 50 range. With the default capacity of 100,000 and the default
+expansion rate of 2, the total items inserted (across all filters) would exceed the 64-bit unsigned integer limit before
+reaching 50 filters.
+
+
+### Choice of Bloom Filter Library
```
parthpatel, line 239: Is the memory for this implementation managed by the same memory allocator as Redis? Or is it separate?

Bloom.md
```diff
@@ -0,0 +1,341 @@
...
+
+
+## Specification
+
+### Bloom Filter Command API
+
+The following Bloom Filter commands are supported and their API syntax is compatible with ReBloom:
+BF.ADD <key> <item>
+BF.EXISTS <key> <item>
+BF.MADD <key> <item> [<item> ...]
```
parthpatel, line 286: do we really need MADD? Can't ADD itself support this?

Bloom.md
```diff
@@ -0,0 +1,341 @@
...
+BF.MEXISTS <key> <item> [<item> ...]
```
parthpatel, line 287: sam question as MADD vs ADD

Bloom.md
parthpatel: +1 also, why not BF.SET. We should follow the naming conventions that valkey has today. BF.SET can probably handle both cases, creation of a new filter and adding items.

Bloom.md
parthpatel: a related question: what is the maximum a bloom object can use? Are there any limits in the implementation or design?

parthpatel (commented):
Mostly comments requesting clarifications, and a few suggestions.

Bloom.md
madolson: There used to be a lot of discussion around variadic vs non-variadic commands. Today the return type of BF.ADD is an integer reply in RedisBloom, whether or not the item was inserted. BF.MADD is an array of integers. We could alter the return type of BF.ADD to be an array (but then it wouldn't be API compatible with RedisBloom) but also then makes the API kind of weird since you are given an array of one element in many cases.

parthpatel: Can't we be smart about it? If there is only 1 return value - do non-array otherwise array. If clients want to expose it as array only, they cast the single object to array. If this is already a closed topic, we can skip rehashing the discussion here.

Bloom.md
KarthikSubbarao: >Could we instead just have a BF.LOAD command?

@madolson The "AOF Rewrite handling" has been updated to explain the usage of BF.LOAD during AOF Rewrite.

>a related question: what is the maximum a bloom object can use? Are there any limits in the implementation or design?

@parthpatel This is explained more in the "Large BloomFilter objects" section

Bloom.md
KarthikSubbarao: > I just realized these are curly quotes. Probably fine to use them, but it would make sense to keep this all as ascii.

Makes sense. This has been removed

Bloom.md
KarthikSubbarao: I updated the section to explain that we will use a fixed seed

Bloom.md
KarthikSubbarao: I agree. I have removed this paragraph and now this section only says we are not compatible with ReBloom & explains why.

Bloom.md
KarthikSubbarao: No, there should not be anything to implement specifically. There is a generic Module keyspace notification system (and APIs) that we can utilize to publish these events.

Bloom.md
KarthikSubbarao: For now, these are the only metrics. In addition to this, I have also explained that the BF.INFO commands can be used to view statistics / info on the bloom objects itself.

I will leave this comment as unresolved incase others would like any other specific metric to be tracked

Bloom.md
KarthikSubbarao: Yes, I updated the section

Bloom.md
KarthikSubbarao: We would expect the same default `bf.bloom_fp_rate` value to be set across the cluster. If this is not the case, it is more likely that adding of an item on the primary succeeds and does not succeed on the replica (false positive indicating that the item already exists).   

Through commands, it is possible to specify the false positive rate through the BF.INSERT and BF.RESERVE commands. BF.ADD and BF.MADD only accept items as arguments. If a user directly starts using BF.ADD, then it will be created using the default false positive rate.

This is fine assuming the user is OK with the default properties. A config will help to override this on a globally for the node. But I understand the concern and also understand the alternative (using BF.INSERT/BF.RESERVE)

Bloom.md
madolson: No, clients hate having variadic return types. Then you have to write custom client code to handle.

> If this is already a closed topic, we can skip rehashing the discussion here.

We used to say the only thing consistent about Redis is that it was inconsistent. I'm just trying to pass on context for decisions, feel free to argue with them :)

Bloom.md
KarthikSubbarao: rebloom does not create and add ACL Categories on the Module commands. This was before the Module ACL Category API was available

Bloom.md
KarthikSubbarao: I can explain this in a Section below. But there is nothing special (in addition to what is already handled by the core).

The core will reject `MODULE UNLOAD` commands if a Module has created a Module Data Type.

Since the ValkeyBloom module creates the bloom data type, `MODULE UNLOAD` will be rejected

Bloom.md
KarthikSubbarao: > how do we serialize the bitmap? 

It is serialized in a byte array format

> what about hash functions?

Hash functions will not be stored in the RDB. Instead, we are using a global static seed from which hash functions are created.

Bloom.md
KarthikSubbarao: >is it the actual operation with original data which replica has to re-hash, or just the location of the bits set as part of the operation?

It is the actual operation itself. The command itself is replicated verbatim and the replica will have to hash the item and set the bits if it does not exist (this check of existence has the configured false positive rate)

madolson:
@valkey-io/core-team @valkey-io/contributors I think this is in a pretty good state to review.

Bloom.md
KarthikSubbarao: When an AOF Rewrite operation is ongoing, the Valkey core will handle this by calling `rewriteAppendOnlyFileRio`. This will loop through items and if the item is of a Module data type, we call its aof_rewrite callback. When this is called, the ValkeyBloom Module will write a BF.LOAD command (with the TTL and the serialized data) into the AOF.

https://github.com/valkey-io/valkey/blob/103365fe0e37b91be20992afd047ba4e3aa3ad47/src/aof.c#L2252

Bloom.md
KarthikSubbarao: I have updated the section to list all the APIs and also list the responses.

Bloom.md
KarthikSubbarao: This refers to the number of items the user wants to add to a bloom object - the add operations itself is based on the configured false positive rate - because add operations can be rejected if we detect (check is based on false positive rate) that the item exists already.

Bloom.md
KarthikSubbarao: This refers to the number of filters within a bloom object. For non scaling, it is always 1. For scaling filters, it is number of scale out operations (each results in creation of a new filter). Each bloom object has a vector of filters within it

Bloom.md
KarthikSubbarao: The number of hash functions depends on the false positive rate. The current library that is used by ValkeyBloom uses a SipHash13 library for hash functions.

Rust Bloom Filter library calculates the number of hash functions needed for the requested false positive rate here: https://github.com/jedisct1/rust-bloom-filter/blob/master/src/lib.rs#L220C8-L220C21

Bloom.md
KarthikSubbarao: A false positive % of 0.1% means 1 in 1000 operations can be "false" and the error rate should be set to 0.001 (Based on ReBloom library's definition)

Bloom.md
KarthikSubbarao: This is one source: https://en.wikipedia.org/wiki/Bloom_filter#Probability_of_false_positives

Bloom.md
enjoy-binbin: i see in json we are doing this:
```
Upon loading, the module registers a new JSON data type. All operations such as query, insert, update and delete are 
efficiently performed on the in-memory document objects, as opposed to JSON text.
* Module name: json
* JSON data type name: ReJSON-RL (Note: We use the same name as the one in RedisJSON for the sake of RDB compatibility.)
```

KarthikSubbarao: We can also consider retaining the same Module name as ReBloom by renaming this to `bf`. 

In case any client uses the `INFO MODULES` or `HELLO` command, they will see the same name.

KarthikSubbarao: @enjoy-binbin @hpatro - Updated the section.

The Module name is "bf" to be compatible with HELLO, INFO, MODULE LIST operations.

The current Module data type name is just "bloom" (different from ReBloom) because it is not RDB compatible with ReBloom.

Bloom.md
```diff
@@ -68,8 +68,9 @@ We have the following terminologies / properties:
 Upon loading, the module registers a new bloom filter module based data type, creates bloom filter (BF.*) commands,
 bloom specific configurations and the bloom ACL category.
 
-* Module name: bloom
+* Module name: bf
```
zuiderkwast, line 71: Can you describe why it's good to use the same module name `bf` as ReBloom? Compatibility for clients that check the HELLO, INFO, etc.

Bloom.md
zuiderkwast: Can you add this explanation in the RFC?

Bloom.md
KarthikSubbarao: Yes, I updated the section with the explanation

Bloom.md
```diff
@@ -0,0 +1,604 @@
+---
+RFC: 4
+Status: Proposed
```
madolson, line 3:
```suggestion
Status: Accepted
```

madolson: Based on the last meeting, this was accepted we were just waiting for more comments.

madolson (approved):

Bloom.md
```diff
@@ -0,0 +1,604 @@
...
+| Total Capacity | Filters per Bloom Object | p50 AVG   | p95 AVG   | p99 AVG   | TPS AVG       |
+|----------------|----------------------|-----------|-----------|-----------|----------------|
+| 1000           | 1                    | 0.263     | 0.30833   | 0.559     | 95218.58667   |
+| 10000          | 10                   | 0.26033   | 0.311     | 0.57233   | 94765.19333   |
+| 25000          | 25                   | 0.255     | 0.30033   | 0.55367   | 98458.69667   |
+| 50000          | 50                   | 0.26833   | 0.50833   | 0.831     | 99669.16667   |
+| 100000         | 100                  | 0.487     | 0.711     | 0.99367   | 83829.30333   |
+| 250000         | 250                  | 0.96167   | 1.17233   | 1.30033   | 49318.69      |
+| 500000         | 500                  | 1.991     | 2.31633   | 2.40167   | 25476.69333   |
+| 1000000        | 1000                 | 4.45233   | 4.76433   | 4.879     | 12034.26667   |
```
madolson, line 285:
```suggestion
| Total Capacity | Filters per Bloom Object | p50 AVG   | p95 AVG   | p99 AVG   | TPS AVG       |
|----------------|--------------------------|-----------|-----------|-----------|---------------|
| 1000           | 1                        | 0.263     | 0.30833   | 0.559     | 95218.58667   |
| 10000          | 10                       | 0.26033   | 0.311     | 0.57233   | 94765.19333   |
| 25000          | 25                       | 0.255     | 0.30033   | 0.55367   | 98458.69667   |
| 50000          | 50                       | 0.26833   | 0.50833   | 0.831     | 99669.16667   |
| 100000         | 100                      | 0.487     | 0.711     | 0.99367   | 83829.30333   |
| 250000         | 250                      | 0.96167   | 1.17233   | 1.30033   | 49318.69      |
| 500000         | 500                      | 1.991     | 2.31633   | 2.40167   | 25476.69333   |
| 1000000        | 1000                     | 4.45233   | 4.76433   | 4.879     | 12034.26667   |
```

Bloom.md
madolson: The explanation was added, but the name was updated to `bloomfltr` btw.

madolson:
Going to merge this as per our discussion last week. Comments can still be made and we can attempt to address them on future updates to the RFC.

Bloom.md
```diff
@@ -0,0 +1,604 @@
...
+0 indicates the item already exists (based on false positve rate).
+```
+(integer) 1
+(integer) 1
+(integer) 0
+```
+
+The following are NEW commands which are not included in ReBloom:
+
+**`BF.LOAD <key> <serialized-value-dump> <TTL>`**
```
KarthikSubbarao, line 524: TTL will not be provided in the command since this will be added to the AOF automatically on the items which have a TTL set. I will update the RFC

PR #6, opened 2024-08-09
Reference RFC file

hpatro:
Reference RFC file for developers to use it as a base and cover all the scenario we would like them to consider for a feature/proposal.


hpatro:
I'm not able to add folks to reviewers. Is there any setting we need to change for this repository @madolson ?

@valkey-io/core-team PTAL

EXAMPLE.md
```diff
@@ -0,0 +1,82 @@
+
+---
+RFC: <PR number>
```
madolson, line 3: I like this, can we attach a reference to the README so the users know they can copy this?

EXAMPLE.md
```diff
@@ -0,0 +1,82 @@
...
+
+(if any)
+
+### Replication
+
+(if any)
+
+### Networking
+
+(if any)
```
madolson, line 58: What does networking mean?

hpatro: Client/protocol changes.

PingXie:
so this is meant to be used as a template for RFS? I like the idea of having a common format. 

Tactical questions:

How do you envision sections like `Commands`, `ACL`, `Cluster`, etc to be filled out? Can you provide some guidance? such as what you expect the author to write about and whether the section is optional, etc.

EXAMPLE.md
madolson: Maybe we should add a one sentence explanation of each of these sections. Might also address Ping's concern below?

hpatro:
@madolson / @PingXie Thanks for the review, added small explanation to each section and linked it from README.md. PTAL.

README.md
```diff
@@ -37,16 +37,4 @@ be referred to by their respective pull-request numbers.
...
-* Abstract. A few sentences describing the feature.
-* Motivation. What the feature solves and why the existing functionality is not
-  enough.
-* Design considerations. A description of the design constraints and
-  requirements for the proposal. Comparisons with similar features in other
-  projects.
-* Specification. A more detailed description of the feature, including why
-  certain details in the design have been chosen.
-* Links to related material such as issues, pull requests, papers, or other references.
+The Valkey RFC format is not a strict format, but should include details about various aspect of the engine, unless they are unnecessary for the proposal you are submitting. Here's the [reference RFC template](REFERENCE.md) to get started.
```
madolson, line 40: Please reflow this so it's consistent in style with the rest of the file.

REFERENCE.md
```diff
@@ -0,0 +1,90 @@
...
+
+If any new commands are introduced:
+
+1. Command name
+   - **Request**
+   - **Response**
+
+### Authentication and Authorization (Optional)
+
+If there are any changes around authentication and authorization (ACL) access for commands/keyspace/channels.
```
madolson, line 34: I would omit this, it seems very niche and not very likely to come up.

REFERENCE.md
```diff
@@ -0,0 +1,90 @@
...
+
+If there are any configuration changes introduced to enable/disable/modify the behavior of the feature.
+
+### Keyspace notifications (Optional)
+
+If there are any events to be introduced or modified to observe activity around the dataset.
+
+### Cluster mode (Optional)
+
+If there are any changes introduced for cluster mode related activities like cluster bus protocol change, slot migration, etc.
```
madolson, line 50: I thought this was about special considerations around clustermode. I think if there is something about changing clustermode, it would likely not be in this section.

REFERENCE.md
```diff
@@ -0,0 +1,90 @@
...
+
+### Module API (Optional)
+
+If any module API is modified or introduced to support dynamic libraries.
```
madolson, line 54:
```suggestion
If any new module APIs are needed to implement or support this feature.
```

REFERENCE.md
```diff
@@ -0,0 +1,90 @@
...
+
+If there are any benchmarks performed and preliminary results are available to share or a set of scenarios identified to measure the feature's performance.
+
+### Testing (Optional)
+
+If there are any test scenarios planned out to support the feature's stability and validate behavior.
+
+### Debug mechanism (Optional)
+
+If there is any debug mechanism introduced to support admin/operators for further maintaining the feature.
```
madolson, line 86:
```suggestion
If there is any debug mechanism introduced to support admin/operators for maintaining the feature.
```

madolson (commented):
Really good, just some minor suggestions.

REFERENCE.md
madolson: Oh, I might have just been confused then. I would just add a description that this includes adding ACL categories as well. I thought you were referring to the overall auth system.

roshkhatri (approved):
This looks pretty good to me too.

REFERENCE.md
```diff
@@ -0,0 +1,90 @@
...
+
+### Networking (Optional)
+
+If there are any changes introduced in the RESP protocol (RESP), client behavior, new server-client interaction mechanism (TCP, RDMA), etc.
+
+### Dependencies (Optional)
+
+If there are any new dependency libraries required to support the feature. Existing dependencies are jemalloc, lua, etc.
+
+### Metrics (Optional)
```
ranshid, line 72: Maybe the 'Metrics' naming does not frame the requirement correctly? 
I would suggest to add a section for 'Observability' in which we can state the different new 'stats', info sections/fields etc... used for both debugging and measurements

REFERENCE.md
```diff
@@ -0,0 +1,90 @@
...
+
+1. Command name
+   - **Request**
+   - **Response**
+
+### Authentication and Authorization (Optional)
+
+If there are any changes around introducing new ACL command/categories for user access control.
+
+### Append-only file (Optional)
```
ranshid, line 36: I agree there might be different considerations about AOF and RDB. but maybe for the sake of simplicity we can set a section name for 'Persistency' which will capture all of these?

ranshid (commented):
@hpatro Really great initiative! 
I think I am missing a general section describing backward compatibility and breaking changes. I think in general these can alternatively be part of the cluster/replicaiton/rdb etc... sections but in case we would like it to be spread around different sections maybe we should explicitly call it out in the sections descriptions?

ranshid:
Another point of thought is describing the external client experience of new features. I like having developers think about how should clients support new features and what is the opt-in mechanism. For example will there be any client SPECs negotiation? and will this new feature require clients to implement client side logic?

REFERENCE.md
```diff
@@ -0,0 +1,90 @@
...
+
+If there are any configuration changes introduced to enable/disable/modify the behavior of the feature.
+
+### Keyspace notifications (Optional)
+
+If there are any events to be introduced or modified to observe activity around the dataset.
+
+### Cluster mode (Optional)
+
+If there is any special handling for this feature (e.g., client redirection, Sharded PubSub, etc) in cluster mode, list out the behavior.
```
PingXie, line 50: I think it would be worthwhile calling out the changes to cluster messages as well, including new extensions, and new types, etc. We have seen compat issues in this area in the past.

REFERENCE.md
```diff
@@ -0,0 +1,90 @@
...
+
+If there are any changes required in the replication mechanism between a primary and replica.
+
+### Networking (Optional)
+
+If there are any changes introduced in the RESP protocol (RESP), client behavior, new server-client interaction mechanism (TCP, RDMA), etc.
+
+### Dependencies (Optional)
+
+If there are any new dependency libraries required to support the feature. Existing dependencies are jemalloc, lua, etc.
```
PingXie, line 70: and whether the dependencies will be vendored (not encouraged), and if yes, explain the reasoning

REFERENCE.md
PingXie: +1 on observability for the title. I don't have a strong opinion for the text (metrics vs stats) other than being consistent with our official documentation (stats?)

REFERENCE.md
```diff
@@ -0,0 +1,90 @@
...
+
+### Metrics (Optional)
+
+If there are any new metrics to be introduced to observe behavior or measure the performance of the feature.
+
+### Benchmarking (Optional)
+
+If there are any benchmarks performed and preliminary results are available to share or a set of scenarios identified to measure the feature's performance.
```
PingXie, line 78: It is also super important to describe how the test is setup, both hardware and software, in addition to the performance numbers

REFERENCE.md
```diff
@@ -0,0 +1,90 @@
...
+
+### Testing (Optional)
+
+If there are any test scenarios planned out to support the feature's stability and validate behavior.
```
PingXie, line 82:
```suggestion
If there are any test scenarios planned to ensure the feature's stability and validate its behavior.
```

PingXie (commented):
LGTM overall. Just some nitpicks. I think we can merge it soon. Let's put it in action and fine-tune it based on the real world feedback (from the RFC authors). Thoughts?

PingXie:
> Another point of thought is describing the external client experience of new features. I like having developers think about how should clients support new features and what is the opt-in mechanism. For example will there be any client SPECs negotiation? and will this new feature require clients to implement client side logic?

I feel that this belongs to the `command` section? but I do agree that we should describe both the syntax and semantics of any new commands introduced.

README.md
```diff
@@ -37,16 +37,6 @@ be referred to by their respective pull-request numbers.
...
-  enough.
-* Design considerations. A description of the design constraints and
-  requirements for the proposal. Comparisons with similar features in other
-  projects.
-* Specification. A more detailed description of the feature, including why
-  certain details in the design have been chosen.
-* Links to related material such as issues, pull requests, papers, or other references.
+The Valkey RFC format is not a strict format, but should include details about
+various aspect of the engine, unless they are unnecessary for the proposal you 
+are submitting. Here's the [reference RFC template](REFERENCE.md) to get started.
```
zuiderkwast, line 42: If the file is a template, I think it should be named TEMPLATE.md rather than REFERENCE.md.

I'm skeptical to deleting this overview description from the README. It provides a short overview that fits within the eyes of the reader without scrolling. Besides, I want to avoid a too formal process with a too strict mandatory format.

A common problem with templates is that people tend to leave a lot of template junk instead of deleting them, like the optional sections.

For these reasons, I think the template should be just an example/suggestion and not a mandatory thing to copy and edit.

PingXie:
> LGTM overall. Just some nitpicks. I think we can merge it soon. Let's put it in action and fine-tune it based on the real world feedback (from the RFC authors). Thoughts?

Almost forgot. I think it will be good to have a version or revision field in this doc as well since we anticipate changes in the future. Otherwise it will be confusing why some RFCs look non conforming after a while.

README.md
hpatro: I started off with EXAMPLE.md but thought REFERENCE might be apt. I'm fine with `TEMPLATE.md` as well. Agaub this is just setting up devs to think through all the scenarios and don't need to conform to this.

REFERENCE.md
hpatro: Observability seems apt. 👍

REFERENCE.md
hpatro: Moving them next to each other.

hpatro:
Thanks @ranshid / @zuiderkwast / @PingXie / @madolson for the review. Have addressed the feedback. PTAL once again.

hpatro:
> LGTM overall. Just some nitpicks. I think we can merge it soon. Let's put it in action and fine-tune it based on the real world feedback (from the RFC authors). Thoughts?

Agreed, I think we can evolve this as we see fit.

hpatro:
> @hpatro Really great initiative! I think I am missing a general section describing backward compatibility and breaking changes. I think in general these can alternatively be part of the cluster/replicaiton/rdb etc... sections but in case we would like it to be spread around different sections maybe we should explicitly call it out in the sections descriptions?

My thought was the RFC author would call out in section's where there is a breaking change. Have left it for now, can always revisit and update the template.

README.md
```diff
@@ -37,16 +37,6 @@ be referred to by their respective pull-request numbers.
...
-* Status and RFC number (the pull-request number).
-* Abstract. A few sentences describing the feature.
-* Motivation. What the feature solves and why the existing functionality is not
-  enough.
-* Design considerations. A description of the design constraints and
-  requirements for the proposal. Comparisons with similar features in other
-  projects.
-* Specification. A more detailed description of the feature, including why
-  certain details in the design have been chosen.
-* Links to related material such as issues, pull requests, papers, or other references.
```
zuiderkwast, line 52: Repeating my previous comment: I don't want to delete these bullets and make the template the single source of information. If we do that, it effectively means that everyone has to use the template.

TEMPLATE.md
```diff
@@ -0,0 +1,91 @@
+---
+RFC: <PR number>
+Status: <Proposed>
+Version: <version number>
```
zuiderkwast, line 4: Version? How do we assign version numbers?

If the file has been updated by multiple PRs, I think we can simply mention those PRs too, so readers can find the discussion on those changes.

The README already mentions this: "For larger changes, the PR making the change is mentioned too and can be referred to by their respective pull-request numbers."

zuiderkwast: So, I don't think we should add version. Instead the RFC field can contain stuff like `RFC: #xxx, updated by #yyy` or possibly we add a History or Updated-by field, but only for the RFCs that have updates.

README.md
hpatro: ack. let me put that back.

TEMPLATE.md
hpatro: I've seen version in few of the tooling system I've used for document/code. We do have git history here so might serve a hint for someone to look at the past. 

@PingXie Do you have anything to add here ?

zuiderkwast: People might not now what to write in a Version field. 1 or 2 or 1.1 or a date?

The PEP process has a *Created* date and a *History* with dates of changes. That's an idea that doesn't need much further documentation. Here's an example: https://peps.python.org/pep-0001/

PingXie: >People might not now what to write in a Version field. 1 or 2 or 1.1 or a date?

My main motivation is to understand why the two RFCs have different formats and I know we will change the formats at some point. 

Here is how this RFC could be used in the future:

I am about to write an RFC for feature X. I make a copy of the latest version of  this RFC, which already comes with a version without any input from me. I then fill in the sections that are applicable. done.

Not sure at which step will the author need to worry/reason about this version field?

PingXie: btw, the exact version scheme is not my concern. We can do semver or just a date like 20240819. As long as we stick to the same version scheme, I am all good.

PingXie: Ah, I see where the confusion is coming from :). Sorry, I meant to say the "meta" version - the version of this RFC template not the real RFC's version. The author can maintain a revision history for their additions/removals/modifications. I think there are a plenty of good systems out there for maintaining revision history. A simple one like "<date>: <change>" would work for me.

PingXie, line 4: I meant to say something like this :) (can be a date or anything else that identifies this particular instance of the template)
```suggestion
Version: 1.0
```

PingXie (commented):
added some more clarification for my earlier comments on "version"

TEMPLATE.md
hpatro: I like the PEP created date / history concept. We can apply it in the future to an actual RFC. @PingXie Do we want this meta version thing ? I'm not sure anymore.

PingXie: I like the PEP format too. I can still see a use of the meta version but not a hard blocker for me.

PingXie (approved):
Thanks!

hpatro:
I think we can update it over time as we see fit. Shall we get it merged @PingXie / @madolson / @zuiderkwast with the current state ?

zuiderkwast (approved):

README.md
```diff
@@ -50,3 +50,5 @@ sections unless they are unnecessary for the proposal you are submitting.
 * Specification. A more detailed description of the feature, including why
   certain details in the design have been chosen.
 * Links to related material such as issues, pull requests, papers, or other references.
+
+Here's a [RFC template](TEMPLATE.md) to get started.
```
madolson, line 54:
```suggestion
Here's an [RFC template](TEMPLATE.md) to get started.
```

madolson (approved):

README.md
madolson, line 54:
```suggestion
Here's an [RFC template](TEMPLATE.md) to get started.
```

PR #11, opened 2024-10-18
Valkey Proxy Proposal

hwware:


ValkeyProxy.md
```diff
@@ -0,0 +1,112 @@
...
+RFC: 10
+Status: <Proposed>
+---
+
+# ValkeyProxy RFC
+
+## Abstract  (Required)
+
+The proposed ValkeyProxy is to provide the features similar to redis-cluster-proxy. 
+
```
vongosling, line 11: aim to decouple xxxx

vongosling (approved):

ValkeyProxy.md
```diff
@@ -0,0 +1,112 @@
...
+        It greatly simplifies the programming complexity of the client, allowing the client to focus more on business logic without caring too much about changes in the underlying architecture.
+
+5.      Performance optimization
+
+        We can use Proxy to optimize in different ways for different clients. For example, we can support Pipeline technology on the Proxy side to improve the overall system response efficiency.
+
+        For some time-consuming or high CPU cost commands, we can make statistics in Proxy side, and for some results, such as min/max/avg/p95/p99 from memtier_benchmark or valkey_benchmark, clients can get timely statistical information.
+
+
+6.	Support audit log
```
vongosling, line 59: Cool, this is a hook in my opinion

ValkeyProxy.md
```diff
@@ -0,0 +1,105 @@
...
+
+        Cloud providers or system administrators can utilize this information to perform rich log statistical summary and 
+correlation analysis functions to achieve a comprehensive audit of information system logs.
+
+7.      Provide reading/writing separation features
+
+        In production environments, most scenarios involve more reading requests and less writing requests. By enabling
+read-write separation feature in the Proxy, we can let the primary node focus on processing write operations and let the 
+replica handle read operations. It is very suitable for scenarios with relatively large read operations and can reduce the 
+pressure on the primary.
```
madolson, line 73: Ooh, this can go horribly wrong. We talked with a customer that was doing caching and did read-write splits, they were having cache misses on replicas but the writes were going to the primary, but the replica was slow at consuming so it would miss again and trigger another write. I think generally read/write splitting is a big anti-pattern for online transactional workloads like caching or message queues. it makes more sense for analytic workloads, but I don't Valkey really serves much analytics today. Maybe that will change

madolson (commented):
I'm fairly dubious of proxies in Valkey as a general solution. They do solve problems at scale related to number of connections, but I think many of the problems can also be solved by improving clients.

There are some cool things that could be solved though, like implementing meta-commands (like distributed locks, sorted sets, etc) which is also very interesting. This isn't exactly the same as your proposal, since you seem to be proposing almost API compatibility with the core.

ValkeyProxy.md
hwware: On benefit implemented in proxy side is to retrieve this information from one point, user do not need to fetch the log in different nodes in the cluster mode.

ValkeyProxy.md
hwware: The reason we want to implement this feature in ValkeyProxy is: 
1. This feature (read-write separation) could be enabled or disabled through config parameter.  And by integrating with ACL, this feature only applies to some specific users
2. For those data with a relatively low update frequency, read misses due to differences in primary and replica data, which in turn lead to multiple writes, are acceptable in some scenarios.
3. If data consistency guarantees are a must following the reason 2 case, synchronous between the master and replica can be considered.

ValkeyProxy.md
```diff
@@ -0,0 +1,105 @@
...
+2.	Shield the functional/API differences between cluster mode and primary-replica mode, so that clients do not need to care about the underlying operating mode when using it.
+
+        Now, If the client is directly connected to the server, then in primary-replica mode and cluster mode, the client may need to execute different code logic for the two modes in the following scenarios:
+        a. The client needs to save sharding metadata information in cluster mode.
+        b. When failover happens in these 2 different modes, the client needs to do different processing.
+        c. In cluster mode, if data migration occurs, the client must handle it.
+
+        So if the client is only connected to the Proxy instead of server node, then as long as the client is not connected inSentinel mode, then when the client sends a command, it does not need to pay too much attention to the underlying connection mode. The Proxy will handle the execution of different modes for the client.
+
+3.	Support multi-database mode even server runs as cluster mode.
```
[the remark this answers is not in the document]
hwware: yes, we can have it in the server side.

ValkeyProxy.md
```diff
@@ -0,0 +1,105 @@
...
+c. When the primary node processes a large key, it may fail to respond to other nodes for a long time, resulting in disconnection of the client.
+
+d. When the number of connections to a node exceeds the maximum value, the client may not be able to connect to the primary node.
+
+Therefore, we need a proxy to make the customer unaware of the above problems, and also to provide some additional Valkey services.
+
+
+## Design Considerations  (Required)
+
+1.	Support all existing Valkey commands.
```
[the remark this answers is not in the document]
hwware: Even not all, but most is necessary, IMO

madolson (commented):
From weekly meeting @PingXie, should we support cross-slot operations (including cross shard operations) . Will we support that functionality and what type of guarantee will provide with that.

ValkeyProxy.md
```diff
@@ -0,0 +1,112 @@
...
+2.	Shield the functional/API differences between cluster mode and primary-replica mode, so that clients do not need to care about the underlying operating mode when using it.
+
+        Now, If the client is directly connected to the server, then in primary-replica mode and cluster mode, the client may need to execute different code logic for the two modes in the following scenarios:
+        a. The client needs to save sharding metadata information in cluster mode.
+        b. When failover happens in these 2 different modes, the client needs to do different processing.
+        c. In cluster mode, if data migration occurs, the client must handle it.
+
+        So if the client is only connected to the Proxy instead of server node, then as long as the client is not connected inSentinel mode, then when the client sends a command, it does not need to pay too much attention to the underlying connection mode. The Proxy will handle the execution of different modes for the client.
+
+3.	Support multi-database mode even server runs as cluster mode
```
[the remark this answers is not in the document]
hwware: Please ref https://github.com/valkey-io/valkey/issues/1319, we have plan to implement multiply db in server side.

ValkeyProxy.md
```diff
@@ -0,0 +1,105 @@
...
+d. When the number of connections to a node exceeds the maximum value, the client may not be able to connect to the primary node.
+
+Therefore, we need a proxy to make the customer unaware of the above problems, and also to provide some additional Valkey services.
+
+
+## Design Considerations  (Required)
+
+1.	Support all existing Valkey commands. 
+
+2.	Shield the functional/API differences between cluster mode and primary-replica mode, so that clients do not need to care about the underlying operating mode when using it.
```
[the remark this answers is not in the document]
hwware: Addressed.

PingXie:
Thank @hwware for putting together this proposal. 

> From weekly meeting @PingXie, should we support cross-slot operations (including cross shard operations) . Will we support that functionality and what type of guarantee will provide with that.

My primary concern here is that simulating standalone behavior on a cluster creates a third mode that is neither standalone nor cluster. Losing atomicity is one of the differences, and then there are pub/sub, streams, scan, eval, functions, etc., to consider. I feel we might end up creating a project that is either 10x more complex or has a much narrower scope (no atomicity, no pub/sub, etc.).

Networking overhead also plays a huge role in in-memory caching performance. For KV caching workloads, where compatibility is less of a concern, adding a network hop would have a very significant impact.

Moreover, the motivations outlined in the RFC don’t necessarily require a proxy to deliver. For example, as @madolson pointed out, Glide already addresses some of these needs with the added potential for end-to-end observability and audit logging.

It seems like the real need here might be a better cluster client rather than a proxy layer. A robust client could solve many of these challenges without introducing unnecessary complexity or performance trade-offs.

ValkeyProxy.md
```diff
@@ -0,0 +1,91 @@
...
+d. When the number of connected clients reaches the 'maxclient' value, the client may not be able to connect to the server.
+
+Therefore, we need a proxy to make the client unaware of the above problems, and also provide some additional Valkey services.
+
+
+## Design Considerations  (Required)
+
+1.	Support most Valkey commands. 
+
+2.	Shield the functional/API differences between cluster mode and standalone mode, so that clients do not need to care about the underlying operating mode when using it.
```
zuiderkwast, line 32: How should we handle things like:

* EVAL scripts that don't declare any keys. Send to all nodes or to a random node? For example, a script that takes a cursor, uses SCAN to find some keys and deletes them and returns the new cursor.
* Cross-slot commands, like MGET, MSET, DEL? Options: 1) Split them per slot and make it non-atomic or 2) reject and return a cross-slot error.
* How to handle cross-slot and cross-shard SUNION, SUNIONSTORE, ZUNION, etc. Return error?

I guess the applications will also need some way to connect directly to the cluster without the proxy, so they can decide for themselves how to do some special commands like the above mentioned ones.

ValkeyProxy.md
```diff
@@ -0,0 +1,91 @@
...
+
+        Now, If the client connects to the server directly, then in standalone mode and cluster mode, the client may need to call different code logic in the following scenarios:
+        a. The client needs to save sharding metadata information in cluster mode.
```
zuiderkwast, line 35: Indentation 4 spaces means it's a code block. It's not correct markdown for this text. Also remove the "(Required)" markers in the headings (copied from the template?).

Please review using the "view file" mode. It looks like this:

Image

zuiderkwast (commented):
I am skeptical to a proxy that tries to hide cluster mode servers behind a standalone-style API. The applications need to be aware because they have some very different semantics with cross-slot command and things like that.

However, it can be a good idea to provide a proxy for standalone-to-standalone, just to multiplex multiple clients into one connection or connection pool, or to proxy cluster-to-cluster mode. The proxy can also hide failovers and route the commands to the primary.

ValkeyProxy.md
```diff
@@ -0,0 +1,91 @@
+---
+RFC: 10
```
zuiderkwast, line 2: This is PR #11.

```suggestion
RFC: 11
```

PR #13, opened 2024-10-25
Addresses open question for ValkeyJSON rfc

roshkhatri:
Addressing open question in the RFC - We would set the config `json.max-document-size` to `0` by default , meaning the size of JSON key will be unlimited.

This is based on the observation that RedisJSON, a related project, does not have a built-in size limit, and the core data types in the Valkey system also do not have such a restriction.

### DIFF
--- a/ValkeyJSON.md
+++ b/ValkeyJSON.md
@@ -81,7 +81,7 @@ We added a limit on JSON key size as a good design practice. Without the limit, 
 The ValkeyJSON module will provide a configuration option, json.max-document-size, that allows users to set a size limit for
 JSON key size.
 
-However, the default value of json.max-document-size will be set to -1, meaning the size of JSON key will be unlimited. 
+However, the default value of json.max-document-size will be set to 0, meaning the size of JSON key will be unlimited. 
 This decision is based on the observation that RedisJSON, a related project, does not have a built-in size limit, and 
 the core data types in the Valkey system also do not have such a restriction.
 
--- a/ValkeyJSON.md
+++ b/ValkeyJSON.md
@@ -1,6 +1,6 @@
 ---
 RFC: 2
-Status: Proposed
+Status: Accepted
 ---
 
 # ValkeyJSON RFC
@@ -75,9 +75,15 @@ called KeyTable, which is a sharded hash table storing key tokens and reference 
 
 ### Document Size Limit
 
-Having a limit on json key size is generally considered as a good design. Without the limit, a malicious program can
-repeatedly call "json.arrinsert" or similar commands to make a json key grow forever, leading to OOM. ValkeyJSON will 
-provide a module config - json.max-document-size - to set the size limit.
+We added a limit on JSON key size as a good design practice. Without the limit, a malicious program could repeatedly call 
+"json.arrinsert" or similar commands to make a JSON key grow indefinitely, potentially leading to an out-of-memory (OOM) condition.
+
+The ValkeyJSON module will provide a configuration option, json.max-document-size, that allows users to set a size limit for
+JSON key size.
+
+However, the default value of json.max-document-size will be set to -1, meaning the size of JSON key will be unlimited. 
+This decision is based on the observation that RedisJSON, a related project, does not have a built-in size limit, and 
+the core data types in the Valkey system also do not have such a restriction.
 
 The amount of memory consumed by a JSON document can be inspected by using the `JSON.DEBUG MEMORY` or `MEMORY USAGE` 
 command. `JSON.DEBUG MEMORY <key> [<path>]` can also report the size of a JSON sub-tree.
@@ -906,14 +912,6 @@ Users can subscribe to the JSON events via the standard keyspace event pub/sub. 
 #### Replication
 
 Every JSON write command is replicated to replicas by calling ValkeyModule_ReplicateVerbatim.
-
-## Open Questions
-
-### Regarding the Document Size Limit
-We are considering a review feedback of not enforcing document size limit, because RedisJSON doesn't have the limit and Valkey 
-core data types don't have limits either. There could be two options:
-1. Make the default value of json.max-document-size 0 or -1, meaning unlimited.
-2. Remove the config.
 
 ## References
 


joehu21:
Looks good.

ValkeyJSON.md
```diff
@@ -75,9 +75,15 @@ called KeyTable, which is a sharded hash table storing key tokens and reference
...
-repeatedly call "json.arrinsert" or similar commands to make a json key grow forever, leading to OOM. ValkeyJSON will 
-provide a module config - json.max-document-size - to set the size limit.
+We added a limit on JSON key size as a good design practice. Without the limit, a malicious program could repeatedly call 
+"json.arrinsert" or similar commands to make a JSON key grow indefinitely, potentially leading to an out-of-memory (OOM) condition.
+
+The ValkeyJSON module will provide a configuration option, json.max-document-size, that allows users to set a size limit for
+JSON key size.
+
+However, the default value of json.max-document-size will be set to -1, meaning the size of JSON key will be unlimited. 
+This decision is based on the observation that RedisJSON, a related project, does not have a built-in size limit, and
```
hpatro, line 85: It's a bit contradictory to reference another project and set the value to unlimited but still introduce this configuration. 
How is this any different to string `APPEND` command? Don't we have the same risk over there?

joehu21: It's a compromised approach based on RFC review feedback. I still think json data type should have a bounded size. But there were objections during the JSON RFC review, saying Valkey core data type like hash doesn't have a size limit (although string value has a limit of 500MB). So, the argument is, for consistency with other data types, JSON shouldn't enforce size limit. So, we compromised. Json will have no size limit by default. If you want size limit, you can override json.max-document-size.

joehu21 (approved):
LGTM

ValkeyJSON.md
hpatro: Yeah, my point is why introduce it just for json documents. With this approach, we might end up with multiple such configs to bother about. I think we should think holistically for all data types and not individual cases separately. So, I feel alternative 2 mentioned below (removing the config) is a good way to start.

joehu21: I'm fine with removing the config outright.

madolson: I think it's generally considered a mistake to not have the configs to limit object size. Especially with cluster mode, which can cause significant performance issues with large keys. I think it makes sense to keep the config with a default value of `infinite`. 

@hpatro Do we want to start an issue on the main repo about object limits in parallel though?

madolson:
@valkey-io/core-team Please approve if you have any concerns with this change to finalize the JSON RFC?

madolson (approved):

enjoy-binbin (approved):

ValkeyJSON.md
zuiderkwast: Zero seems much more common than -1 to for unlimited/disabled. (Searching for the word "zero" in valkey.conf, I find the following examples of zero = unlimited: client-output-buffer-limit, tls-session-cache-size, auto-aof-rewrite-percentage, latency-monitor-threshold, stream-node-max-bytes. There are probably more, like the listpack size limits, that aren't even mentioned in the same way.)

Historically, there have been no limits to key sizes, I think the default should be unlimited. I don't mind adding configs though. Huge keys are a bit like KEYS or very heavy Lua scripts. These are also not blocked by default, though it's possible to limit using ACL.

If we add memory tracking to all types, can we set a single config to limit any the size of any value? I don't know if it makes sense.

@joehu21 There is no limit for string keys, but there is a limit for the RESP protocol strings. You can create a huge string using the APPEND command and retrieve parts of it using GETRANGE.

ValkeyJSON.md
```diff
@@ -75,9 +75,15 @@ called KeyTable, which is a sharded hash table storing key tokens and reference
...
-Having a limit on json key size is generally considered as a good design. Without the limit, a malicious program can
-repeatedly call "json.arrinsert" or similar commands to make a json key grow forever, leading to OOM. ValkeyJSON will 
-provide a module config - json.max-document-size - to set the size limit.
+We added a limit on JSON key size as a good design practice. Without the limit, a malicious program could repeatedly call 
+"json.arrinsert" or similar commands to make a JSON key grow indefinitely, potentially leading to an out-of-memory (OOM) condition.
+
+The ValkeyJSON module will provide a configuration option, json.max-document-size, that allows users to set a size limit for
+JSON key size.
+
+However, the default value of json.max-document-size will be set to -1, meaning the size of JSON key will be unlimited.
```
zuiderkwast, line 84:
```suggestion
However, the default value of json.max-document-size will be set to 0, meaning the size of JSON key will be unlimited. 
```

roshkhatri: I think 0 is zero/disabled, `-1` is more logical as to indexes as `-1` also refers to last and here its infinite.

ValkeyJSON.md
joehu888: Understood it's more common to use zero to represent unlimited. The intention of using -1 for unlimited is that we can use 0 as a means to disable write commands. If the limit is set to zero, all json write commands are rejected. But I am also fine with using zero to represent unlimited.

zuiderkwast: Interesting point. I didn't think about this.

But it's also possible to make it read-only using ACL rules, right? Using zero for this seems like hack.

zuiderkwast (approved):

hwware:
I notice after Madolson and BinBin approve, Vikctor update the json.max-document-size as 0, Can we confirm 0 is the correct default value?

ValkeyJSON.md
madolson: >Historically, there have been no limits to key sizes, I think the default should be unlimited. I don't mind adding configs though. Huge keys are a bit like KEYS or very heavy Lua scripts. These are also not blocked by default, though it's possible to limit using ACL.

FYI, we still limit strings to the proto_max_bulk_len even with append. This codebase, https://github.com/valkey-io/valkey/blob/2df56d87c0ebe802f38e8922bb2ea1e4ca9cfa76/src/t_string.c#L40C12-L40C29, is called on the append flow. 

> Understood it's more common to use zero to represent unlimited. The intention of using -1 for unlimited is that we can use 0 as a means to disable write commands. If the limit is set to zero, all json write commands are rejected. But I am also fine with using zero to represent unlimited.

I have a minor preference for zero to indicate infinite. I think you can achieve the same functionality by just saying setting to 1 indicates disabled write commands. There are no valid json blobs that are 1 byte :) (ACL also works)

madolson:
> I notice after Madolson and BinBin approve, Viktor update the json.max-document-size as 0, Can we confirm 0 is the correct default value?

Yeah, I think 0 is the right value.

ValkeyJSON.md
joehu21: agreed

zuiderkwast: > FYI, we still limit strings to the proto_max_bulk_len even with append.

TIL

But DUMP can still return larger bulk replies than `proto_max_bulk_len`, right?

hwware (approved):

ValkeyJSON.md
madolson: > But DUMP can still return larger bulk replies than proto_max_bulk_len, right?

Yeah, we actually have gone in Amazon now that I look to omit this for slot migration, which is the only place that dump is ignored. It's on my checklist for whenever we finalize atomic slot migration to make sure it's incorporated.

PR #8, opened 2024-08-27
RFC for Valkey Vector Search

kyle-meggs:
This proposed Valkey Vector Search module will introduce KNN (“Flat”) and ANN (“HNSW”) vector similarity search algorithms to provide vector search capabilities on both Valkey Cluster and Valkey Standalone. With vector search on Valkey, users can take advantage of low latency in-memory vector search in their familiar Valkey environment.

PR #15, opened 2024-11-27
Revised VSS RFC

allenss-amazon:
Replaces and extends the previous PR for VSS module.

### DIFF
--- a/VSS.md
+++ b/VSS.md
@@ -1,0 +1,415 @@
+---
+
+## RFC: Status: Proposed
+
+
+# Valkey Search RFC
+
+## Abstract
+
+This proposed Valkey Search module will introduce search algorithms to provide numeric, tag and vector search capabilities on both Valkey Cluster and Valkey Standalone. With search on Valkey, users can take advantage of low latency in-memory search in their familiar Valkey environment.
+
+## Motivation
+
+With the rising popularity of generative AI, developers have turned to in-memory data stores such as Valkey or Redis for ultra-low latency (p99 single digit ms) vector search for use cases like Retrieval Augmented Generation (RAG), recommendation systems, semantic search, semantic caching, and more. For developers already using Valkey or Redis, co-locating their vectors and search functionality alongside their underlying data will accelerate and simplify adoption. And for other “greenfield” users, the ultra-low latency of Valkey search may attract them to replace their existing  database for high performance use cases to achieve in-memory speeds.
+
+## Design considerations
+
+Here we list Critical User Journeys (CUJs) which we believe Valkey users will desire as part of a Search Module.
+
+**Search Operations CUJ** \- as a user, I want to perform search operations using commonly available clients.
+
+- Requirement: support index insertion/mutation/deletion/query operations  
+- Requirement: maintain compatibility with RediSearch VSS APIs, Memorystore APIs, and MemoryDB APIs as much as possible
+
+**Performance CUJ** \- As a user, I want to use in-memory search because it provides the lowest latency, and most performant option
+
+- Requirement: low latency data ingestion  
+- Requirement: low latency queries  
+- Requirement: high throughput: query \+ ingestion  
+- Requirement: high throughput at high levels of recall  
+- Requirement: \[Scale up\] multi-threading & vertical scaling (query \+ ingestion)  
+- Requirement: Negligible resource consumption (CPU, memory) for Valkey applications and operations/commands that do not use the search module.
+
+**API Compatibility CUJ** \- As a user, I want a compatible API to RediSearch VSS, which is supported by common clients. The goal is to enable applications to be portable between Redisearch and Valkey search with minimal effort. The search module will strive for the highest level of compatibility with Redisearch API for core application operations that are valid, deterministic and error-free. Other items, e.g., management or configuration commands, timing dependent operations, programmer development commands, status commands and/or error messages for invalid operations may differ substantially.
+
+Binary RDB and replication stream compatibility with Redisearch is not needed, meaning that online/live data migration will not be possible.
+
+**HASH and JSON Data Type support CUJ** \- As a user, I want to store my vectors in a HASH or JSON data type, which is natively supported in Valkey.
+
+**Scale Up & Out CUJ** \- As a user, I want to be able to both scale up and out
+
+- Requirement: multi-threading for scale up  
+- Requirement: Cluster for scale out & storing more data
+
+**Consistency CUJ** \- As a user, I want to be able to select a consistency model that matches my needs.
+
+- Requirement: A sequential consistency model must be supported.
+
+**Memory footprint CUJ** \- As a user, I want the index to consume memory based on usage and dynamically grow
+
+- Requirement: Lazy memory consumption based on the index entries and the dimensions.  
+- Requirement: Index size should be allowed to grow dynamically with minimal performance overhead
+
+**Replication CUJ** \- As a user, I want to rely on replication to provide high-availability and ensure that my data is protected by storing it across multiple nodes with low replication latency
+
+- Requirement: Support fast bootstrap  
+- Requirement: Support index full sync on replication lag.
+
+**Monitoring CUJ** \- As a user, I need to be able to monitor indexes and re-indexing with simple observability metrics
+
+- Requirement: Ability to monitor index base statistics  
+- Requirement: Ability to monitor cross indexes, aggregated, view statistics.
+
+**Backfill CUJ** \- As a user, I need to be able to create an index after the keyspace contains vector entries
+
+- Requirement: Index creation triggers a backfill process which indexes the existing keyspace entries.
+
+**Hybrid Query CUJ** \- As a user, I want to be able to perform hybrid queries to incorporate filtering criteria in addition to best-match vector search
+
+- Requirement: Support for filtering numeric and tag-based filters.  
+- Requirement: Support for boolean combinations of operators in combination with filtering (AND, OR, NOT)
+
+**Result Set Management CUJ** \- As a user I want to dictate which fields are returned in a search
+
+- Requirement: Support query controls to set the desired returned fields.
+
+**ACLs CUJ** \- As a user of Access Control Lists (ACLs), I want to manage user-based permissions, specific command permissions, etc to manage vector search access
+
+- Requirement: Existing ACls functionality should extend to vector search commands
+
+**Persistence CUJ** \- As a user I want to be able to leverage RDB and/or AOF for persisting and loading indexes
+
+- Requirement: Index metadata should be persisted leveraging RDB  
+- Requirement: Index types that take significant time to rebuild, such as vector indexes, must support persistence of the index payload.
+
+**Tuning CUJ** \- As a user who is also managing the infrastructure, I want to tune my configurations, like number of threads, to optimize my search performance
+
+- Requirement: Configs like number of threads should be exposed and tunable to users
+
+## Specification
+
+The search module allows a user to create any number of indexes. Each index consists of any number of typed fields and covers a specified subset of the keys within a Valkey database.
+
+The creation of an index initiates a backfill process in the background that will scan the keyspace and insert all covered keys into the newly created index. Mutations of a key are automatically reflected in the index. This process is automatic after the creation of the index. Mutation of keys is explicitly allowed during the backfill process. The current state of a backfill may be interrogated by a client using the FT.INFO command.
+
+A query operation is a two step process. In the first step, a query is executed that generates a set of keys and distances. In the second step, that list of keys and distances is used to generate a response.
+
+The first step is performed in a background thread, while the second step must be performed by the mainthread. If a key produced by step 1 is mutated before step 2 is executed by the mainthread, the result of the query may not be consistent with the latest state of the keys.
+
+### Commands
+
+```
+FT.CREATE <index-name>
+    ON HASH
+    [PREFIX <count> <prefix> [<prefix>...]]
+    SCHEMA 
+        (
+            <field-identifier> [AS <field-alias>] 
+                  NUMERIC 
+                | TAG [SEPARATOR <sep>] [CASESENSITIVE] 
+                | VECTOR [HNSW | FLAT] <attr_count> [<attribute_name> <attribute_value>]+)
+        )+
+```
+
+The `FT.CREATE` command creates an empty index and initiates the backfill process. Each index consists of a number of field definitions. Each field definition specifies a field name, a field type and a path within each indexed key to locate a value of the declared type. Some field type definitions have additional sub-type specifiers.
+
+For indexes on HASH keys, the path is the same as the hash member name. The optional `AS` clause can be used to rename the field if desired. This is especially useful when the member name contains special characters.
+
+- **\<index-name\>** (required): This is the name you give to your index. If an index with the same name exists already, an error is returned.  
+    
+- **ON HASH** (optional): Only keys that match the specified type are included into this index. If omitted, HASH is assumed.  
+    
+- **PREFIX \<prefix-count\> \<prefix\>** (optional): If this clause is specified, then only keys that begin with the same bytes as one or more of the specified prefixes will be included into this index. If this clause is omitted, all keys of the correct type will be included. A zero-length prefix would also match all keys of the correct type.  
+    
+- Field types:  
+    
+  - **TAG**: A tag field is a string that contains one or more tag values.  
+      
+    - **SEPARATOR \<sep\>** (optional): One of these characters `,.<>{}[]"':;!@#$%^&*()-+=~` used to delimit individual tags. If omitted the default value is `,`.  
+    - **CASESENSITIVE** (optional): If present, tag comparisons will be case sensitive. The default is that tag comparisons are NOT case sensitive
+
+    
+
+  - **NUMERIC** Field contains a number.  
+      
+  - **VECTOR**: Field contains a vector. Two vector indexing algorithms are currently supported: HNSW (Hierarchical Navigable Small World) and FLAT (brute force). Each algorithm has a set of additional attributes, some required and other optional.  
+      
+    - **FLAT:** The Flat algorithm provides exact answers, but has runtime proportional to the number of indexed vectors and thus may not be appropriate for large data sets.  
+      - **DIM \<number\>** (required): Specifies the number of dimensions in a vector.  
+      - **TYPE FLOAT32** (required): Data type, currently only FLOAT32 is supported.  
+      - **DISTANCE\_METRIC \[L2 | IP | COSINE\]** (required): Specifies the distance algorithm  
+      - **INITIAL\_CAP \<size\>** (optional): Initial index size.  
+    - **HNSW:** The HNSW algorithm provides approximate answers, but operates substantially faster than FLAT.  
+      - **DIM \<number\>** (required): Specifies the number of dimensions in a vector.  
+      - **TYPE FLOAT32** (required): Data type, currently only FLOAT32 is supported.  
+      - **DISTANCE\_METRIC \[L2 | IP | COSINE\]** (required): Specifies the distance algorithm  
+      - **INITIAL\_CAP \<size\>** (optional): Initial index size.  
+      - **M \<number\>** (optional): Number of maximum allowed outgoing edges for each node in the graph in each layer. on layer zero the maximal number of outgoing edges will be 2\*M. Default is 16 Maximum is 512\.  
+      - **EF\_CONSTRUCTION \<number\>** (optional): controls the number of vectors examined during index construction. Higher values for this parameter will improve recall ratio at the expense of longer index creation times. The default value is 200\. Maximum value is 4096\.  
+      - **EF\_RUNTIME \<number\>** (optional):  Sets the count of vectors to be examined during a query operation. The default is 10, and the max is 4096\. You can set this parameter value for each query you run. Higher values increase query times, but improve query recall.
+
+**RESPONSE** OK or error.
+
+```
+FT.DROPINDEX <index-name>
+```
+
+The specified index is deleted. It is an error if that index doesn't exist.
+
+- **\<index-name\>** (required): The name of the index to delete.
+
+**RESPONSE** OK or error.
+
+```
+FT.INFO <index-name>
+```
+
+Detailed information about the specified index is returned.
+
+- **\<index-name\>** (required): The name of the index to return information about.
+
+**RESPONSE**
+
+An array of key value pairs.
+
+- **index\_name**	(string)	The index name  
+- **num\_docs**	(integer)	Total keys in the index  
+- **num\_records**	(integer)	Total records in the index  
+- **hash\_indexing\_failures**	(integer)	Count of unsuccessful indexing attempts  
+- **indexing**	(integer)	Binary value. Shows if background indexing is running or not  
+- **percent\_indexed**	(integer)	Progress of background indexing. Percentage is expressed as a value from 0 to 1  
+- **index\_definition**	(array)	An array of values defining the index  
+  - **key\_type**	(string)	HASH. This is the only available key type.  
+  - **prefixes**	(array of strings)	Prefixes for keys  
+  - **default\_score**	(integer) This is the default scoring value for the vector search scoring function, which is used for sorting.  
+  - **attributes**	(array)	One array of entries for each field defined in the index.  
+    - **identifier**	(string)	field name  
+    - **attribute**	(string)	An index field. This is correlated to a specific index HASH field.  
+    - **type**	(string)	VECTOR. This is the only available type.  
+    - **index**	(array)	Extended information about this internal index for this field.  
+      - **capacity**	(integer)	The current capacity for the total number of vectors that the index can store.  
+      - **dimensions**	(integer)	Dimension count  
+      - **distance\_metric**	(string)	Possible values are L2, IP or Cosine  
+      - **data\_type**	(string)	FLOAT32. This is the only available data type  
+      - **algorithm**	(array)	Information about the algorithm for this field.  
+        - **name**	(string)	HNSW or FLAT  
+        - **m**	(integer)	The count of maximum permitted outgoing edges for each node in the graph in each layer. The maximum number of outgoing edges is 2\*M for layer 0\. The Default is 16\. The maximum is 512\.  
+        - **ef\_construction**	(integer)	The count of vectors in the index. The default is 200, and the max is 4096\. Higher values increase the time needed to create indexes, but improve the recall ratio.  
+        - **ef\_runtime**	(integer)	The count of vectors to be examined during a query operation. The default is 10, and the max is 4096\.
+
+```
+FT._LIST
+```
+
+Lists the currently defined indexes.
+
+**RESPONSE**
+
+An array of strings which are the currently defined index names.
+
+```
+FT.SEARCH <index> <query>
+  [NOCONTENT]
+  [TIMEOUT <timeout>]
+  [PARAMS nargs <name> <value> [ <name> <value> ...]]
+  [LIMIT <offset> <num>]
+  [DIALECT <dialect>]
+```
+
+Performs a search of the specified index. The keys which match the query expression are returned.
+
+- **\<index\>** (required): This index name you want to query.  
+- **\<query\>** (required): The query string, see below for details.  
+- **NOCONTENT** (optional): When present, only the resulting key names are returned, no key values are included.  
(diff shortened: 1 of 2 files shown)


madolson (commented):
Other open questions that would document for completeness:
1. Will ACLs be supported?
2. Is active defragmentation supported?
3. Is the RDB compatible with redis search.
4. I assume this a subset of redis search functionality. but would be good to document that along with what is missing.
5. How will slot migration work, if at all. Will we also copy indexes between nodes when using the cli?
6. Are any core changes required? We talked about memory sharing and some new keysspace notifications.
7. How does redirecting read requests to the primary work in cluster mode? Can you force strongly consistent reads? By the consistency model do we care?
8. Are we allowing searcg in Lua and multi-exec? Will it cause latency issues? Does redis allow it?

VSS.md
```diff
@@ -0,0 +1,415 @@
...
+- Requirement: high throughput at high levels of recall  
+- Requirement: \[Scale up\] multi-threading & vertical scaling (query \+ ingestion)  
+- Requirement: Negligible resource consumption (CPU, memory) for Valkey applications and operations/commands that do not use the search module.
+
+**API Compatibility CUJ** \- As a user, I want a compatible API to RediSearch VSS, which is supported by common clients. The goal is to enable applications to be portable between Redisearch and Valkey search with minimal effort. The search module will strive for the highest level of compatibility with Redisearch API for core application operations that are valid, deterministic and error-free. Other items, e.g., management or configuration commands, timing dependent operations, programmer development commands, status commands and/or error messages for invalid operations may differ substantially.
+
+Binary RDB and replication stream compatibility with Redisearch is not needed, meaning that online/live data migration will not be possible.
+
+**HASH and JSON Data Type support CUJ** \- As a user, I want to store my vectors in a HASH or JSON data type, which is natively supported in Valkey.
```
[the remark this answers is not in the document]
allenss-amazon: Not in scope for first release.

VSS.md
```diff
@@ -0,0 +1,415 @@
...
+- **LIMIT \<offset\> \<count\>** (optional): Lets you choose a portion of the result. The first `<offset>` keys are skipped and only a maximum of `<count>` keys are included. The default is LIMIT 0 10, which returns at most 10 keys.  
+- **DIALECT \<dialect\>** (optional): Specifies your dialect. The only supported dialect is 2\.
+
+**RESPONSE**
+
+The command returns either an array if successful or an error.
+
+If `NOCONTENT` is specified, then the output is .....
+
+If `NOCONTENT` is not specified, then the output is an array of (2\*N)+1 entries, where N is the number of keys output from the search. The first entry in the array is the value N which is followed by N pairs of entries, one per key found. Each pair of entries consists of the key name followed by an array which is the result value for that key.
```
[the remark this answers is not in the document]
allenss-amazon: compatibility with Redisearch

VSS.md
```diff
@@ -0,0 +1,415 @@
...
+
+The following query will return all books that either don't have a genre field, or have a genre field not equal to "comedy", that are published between 2015 and 2024:
+
+`-@genre:[comedy] @year:[2015 2024]`
+
+### Configuration 
+
+1\. **reader-threads:** (Integer) Controls the amount of threads executing queries.  
+2\. **writer-threads:** (Integer) Controls the amount of threads processing index mutations.  
+3\. **use-coordinator:** (boolean) Cluster mode enabler.
```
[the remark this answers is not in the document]
allenss-amazon: yes. will add.

yairgott:
> Other open questions that would document for completeness:
> 
> 1. Will ACLs be supported?
ACLs are not supported in the current version. 

> 2. Is active defragmentation supported?
Memory allocated by the module is not subject to active defragmentation.

> 3. Is the RDB compatible with redis search.
No, the RDB format is not compatible with RediSearch. The module utilizes a proprietary RDB format based on protobuf, which serializes both index metadata and content. In contrast, RediSearch's RDB format only serializes the index metadata.

> 4. I assume this a subset of redis search functionality. but would be good to document that along with what is missing.
Yes, we should have a section which enumerates key missing functionality as bullet points to provide clarity. 

> 5. How will slot migration work, if at all. Will we also copy indexes between nodes when using the cli?
Slot migration is supported. Keys from the migrated slot trigger keyspace mutation events, which are processed in the same manner as client-originated mutations, except that they do not block the client during execution. 

> 6. Are any core changes required? We talked about memory sharing and some new keysspace notifications.
While the module can function without engine changes, two engine changes have been introduced to enhance user experience:

1. Blocking client on keyspace mutation events: Ensures visibility of mutations in consecutive queries on the same connection. This change extends the behavior of the module API RedisModule_BlockClient.
2. Memory deduplication: Adds new module APIs to enable memory sharing between the module and the engine, reducing duplication.

> 7. How does redirecting read requests to the primary work in cluster mode? Can you force strongly consistent reads? By the consistency model do we care?
Valkey doesn't support distributed transactions so I don't see any consistency guarantee with scatter-gather across multiple shards.
Additional context: In cluster mode, a query can be received by any node, which fans out the request to all shards. Each shard processes the query locally and returns results to the initiating node, which selects the top responses across shards. Currently, forcing reads from primary nodes only are not supported. The current logic for instance selection is designed to improve the performance, increase throughput and lower latency, by load balancing between the instances. 

> 8. Are we allowing searcg in Lua and multi-exec? Will it cause latency issues? Does redis allow it?
Will ACLs be supported?
* Multi-exec: Supported, but it introduces significant performance impacts. Mutation processing cannot be parallelized, and queries inside a multi-exec context must wait for preceding mutations to complete.
* Lua scripts: Supported, and the client is not blocked during mutation operations triggered by a Lua script.

yairgott:
Made some inline editing to bullet #8 regarding to LUA.

VSS.md
```diff
@@ -0,0 +1,484 @@
...
+With the rising popularity of generative AI, developers have turned to in-memory data stores such as Valkey or Redis for ultra-low latency (p99 single digit ms) vector search for use cases like Retrieval Augmented Generation (RAG), recommendation systems, semantic search, semantic caching, and more. For developers already using Valkey or Redis, co-locating their vectors and search functionality alongside their underlying data will accelerate and simplify adoption. And for other “greenfield” users, the ultra-low latency of Valkey search may attract them to replace their existing  database for high performance use cases to achieve in-memory speeds.
+
+## Design considerations
+
+Here we list Critical User Journeys (CUJs) which we believe Valkey users will desire as part of a Search Module.
+
+**Search Operations CUJ** \- as a user, I want to perform search operations using commonly available clients.
+
+- Requirement: support index insertion/mutation/deletion/query operations  
+- Requirement: maintain compatibility with RediSearch VSS APIs, Memorystore APIs, and MemoryDB APIs as much as possible
```
zhangh43, line 23: RedisSearch API is weird. Pinecone/Milvus API is more natural
```
# create
pinecone.create_index(name=index_name, dimension=dimension)
# insert
index = pinecone.Index(index_name)
index.upsert(vectors)
# search
results = index.query(queries=[query_vector], top_k=5)
```

I believe developers prefer the latter.

madolson: So to clarify the difference:
1. In the current implementation, you insert keys with other commands and then create an index ontop of a prefix structure.
2. In the pinecone world, you create an index and then insert objects into the index explicitly.

?

VSS.md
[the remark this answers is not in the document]
madolson: Yes, I see it's updated below now.

VSS.md
```diff
@@ -0,0 +1,415 @@
...
+2. **Stability test framework:** fuzzy testing based on python.
+
+### Observability 
+
+The following metrics are added to the INFO command.
+
+- **search\_total\_indexed\_hash\_keys**	(Integer)	Total count of HASH keys for all indexes  
+- **search\_number\_of\_indexes**	(Integer)	Index schema total count  
+- **search\_number\_of\_attributes**	(Integer)	Total count of attributes for all indexes  
+- **search\_failure\_requests\_count**	(Integer)	A count of all failed requests
```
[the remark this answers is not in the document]
madolson: Yeah, I think typically those are called errors throughout the codebase.

VSS.md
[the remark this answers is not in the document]
madolson: So, we have no idea why RedisSearch added it?

VSS.md
```diff
@@ -0,0 +1,484 @@
...
+### Testing 
+
+1. **Unittesting:** code coverage currently stands at 90%.  
+2. **Stability test framework:** fuzzy testing based on python.
+
+### Observability 
+
+The following metrics are added to the INFO command.
+
+- **search\_total\_indexed\_hash\_keys**	(Integer)	Total count of HASH keys for all indexes
```
madolson, line 408: Should there be a corresponding JSON variant?

VSS.md
```diff
@@ -0,0 +1,484 @@
...
+- **search\_failure\_requests\_count**	(Integer)	A count of all failed requests, including syntax errors.
+- **search\_successful\_requests\_count**	(Integer)	A count of all successful requests  
+- **search\_used\_memory\_human**	(Integer)	A human-friendly readable version of the search\_used\_memory\_bytes metric  
+- **search\_used\_memory\_bytes**	(Integer)	The total bytes of memory that all indexes occupy  
+- **search\_background\_indexing\_status**	(String)	The status of the indexing process. NO\_ACTIVITY indicates idle indexing.  
+- **search\_hnsw\_create\_exceptions\_count**	(Integer)	Count of HNSW creation exceptions.  
+- **search\_hnsw\_search\_exceptions\_count**	(Integer)	Count of HNSW search exceptions  
+- **search\_hnsw\_remove\_exceptions\_count**	(Integer)	Count of HNSW removal exceptions.  
+- **search\_hnsw\_add\_exceptions\_count**	(Integer)	Count of HNSW addition exceptions.  
+- **search\_hnsw\_modify\_exceptions\_count**	(Integer)	Count of HNSW modification exceptions
```
madolson, line 420: Still unclear, what are exceptions? How do these fails and what are end users supposed to be doing about these. Are these just syntax errors?

VSS.md
```diff
@@ -0,0 +1,415 @@
...
+- **search\_hnsw\_add\_exceptions\_count**	(Integer)	Count of HNSW addition exceptions.  
+- **search\_hnsw\_modify\_exceptions\_count**	(Integer)	Count of HNSW modification exceptions  
+- **search\_modify\_subscription\_skipped\_count**	(Integer)	Count of skipped subscription modifications  
+- **search\_remove\_subscription\_successful\_count**	(Integer)	Count of successful subscription removals  
+- **search\_remove\_subscription\_skipped\_count**	(Integer)	Count of skipped subscription removals  
+- **search\_remove\_subscription\_failure\_count**	(Integer)	Count of failed subscription removals  
+- **search\_add\_subscription\_successful\_count**	(Integer)	Count of successfully added subscriptions  
+- **search\_add\_subscription\_failure\_count**	(Integer)	Count of failures of adding subscriptions  
+- **search\_add\_subscription\_skipped\_count**	(Integer)	Count of skipped subscription adding processes  
+- **search\_modify\_subscription\_failure\_count**	(Integer)	Count of failed subscription modifications
```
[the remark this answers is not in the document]
madolson: Still not sure what these subscriptions are, they are only mentioned here.

VSS.md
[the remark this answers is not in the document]
madolson: Followup, is there a reason to disable the coordinator for cluster mode? Should it just automatically get enabled for cluster mode?

VSS.md
```diff
@@ -0,0 +1,484 @@
...
+
+The first step is performed in a background thread, while the second step is performed by the mainthread. If a key produced by step 1 is mutated before step 2 is executed by the mainthread, the result of the query may not be consistent with the latest state of the keys.
+
+### Cluster Mode Operation
+
+In cluster mode, indexes are sharded just like data, meaning that each node maintains an index only for the keys that are present on that node.
+This means for that for data mutation operations, no cross-cluster communication is required, each node simply updates it's local index to reflect the mutation of the local key.
+
+Query operations are accepted by any node in the cluster and that node is responsible for broadcasting the query across the cluster and merging the results for delivery to the client.
+Cross-client communication uses gRPC and protobufs and does not require mainthread interaction.
```
madolson, line 106: Is this on a separate port? Does this need to be configurable so that end users can slot it into their system?

VSS.md
```diff
@@ -0,0 +1,518 @@
+## RFC: 8
+
+## Status: Proposed
```
zuiderkwast, line 3: This should be markdown metadata (frontmatter) and the RFC number should be the PR number.

```suggestion
---
RFC: 15
Status: Proposed
---
```

VSS.md
```diff
@@ -0,0 +1,518 @@
...
+
+- Requirement: Configs like number of threads should be exposed and tunable to users
+
+## Current Specification
+
+The search module allows a user to create any number of indexes. Each index consists of any number of typed fields and covers a specified subset of the keys within a Valkey database.
+
+The creation of an index initiates a backfill process in the background that will scan the keyspace and insert all covered keys into the newly created index. Mutations of a key are automatically reflected in the index. This process is automatic after the creation of the index. Mutation of keys is explicitly allowed during the backfill process. The current state of a backfill may be interrogated by a client using the FT.INFO command.
+
+A query operation is a two step process. In the first step, a query is executed that generates a set of keys and distances. In the second step, that list of keys and distances is used to generate a response.
```
zuiderkwast, line 96: This section is about how the data is represented and stored. I miss information about how all of this is stored in RDB files, AOF files, replicated to replicas and stored in slot migration.

What are "keys and distances"?

* Are they temporary data just only produce one response and then discarded?
* Are they as actual keys in the database? If not, it should be clarified, because when the word "keys" is used, it can be confusing what it is.
* Are these keys and distances stored in RDB, AOF and replicated?
* Are they sharded or local to one node?

VSS.md
zuiderkwast: This communication needs to be explained in some more depth. How do the nodes exchange the grpc-port with each other, over the cluster bus?

Is there already some module API for modules to use the cluster bus to exchange this kind of information or do we have to add it?

VSS.md
```diff
@@ -0,0 +1,518 @@
...
+
+The first step is performed in a background thread, while the second step is performed by the mainthread. If a key produced by step 1 is mutated before step 2 is executed by the mainthread, the result of the query may not be consistent with the latest state of the keys.
+
+### Cluster Mode Operation
+
+In cluster mode, indexes are sharded just like data, meaning that each node maintains an index only for the keys that are present on that node.
+This means for that for data mutation operations, no cross-cluster communication is required, each node simply updates it's local index to reflect the mutation of the local key.
```
zuiderkwast, line 103: Does this mean that indexes don't need to be replicated? Please elaborate and mention something about these cases:

* Are indexes replicated and stored in RDB and AOF files?
* Do the indexes not need to be replicated, because the replica builds its own index from the keys it receives from the primary?
* The same questions about slot migration: Do the indexes need to be moved with slot migration or does the importing node just build its own index with the keys it receives from the migrating node?

zuiderkwast (commented):
I've read the RFC and it's interesting. I don't know much about vector similarity search, so the introduction and motivation was very useful.

I didn't review the individual commands. I assume those are similar to the redis and other APIs and I don't know enough about it, but I put some comments about the specification, especially the communication with other nodes and data persistence, etc. I think it needs some more details.

I'm also missing some discussion about what happens in case of failover, corner cases like failover during a distributed search, outdated data, etc. whatever can happen. Are there race conditions regarding if keys are modified at the same time as a search is started? What about consistency in a distributed search, are there any guarantees?

Is a global cross-slot search possible or is everything per cluster slot or per shard? Implicitly, it seems to be global search, but I think this should be mentioned because with most of the existing commands, like SUNION, cross-slot operations is not allowed, while in other commands like SCAN, keys from multiple slots can be returned in the same call.

Regarding the gRPC communication, can you explain the RCP calls you're using, if not in detail then at least in some pseudo-code style?

VSS.md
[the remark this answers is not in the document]
zuiderkwast: "JSON data type, which is natively supported in Valkey"

This is not true. There is a JSON module. Can the VSS module use the JSON module, like an optional dependency? Or mandatory dependency? Please elaborate a little.

VSS.md
```diff
@@ -0,0 +1,518 @@
...
+- Requirement: Support for filtering numeric and tag-based filters.  
+- Requirement: Support for boolean combinations of operators in combination with filtering (AND, OR, NOT)
+
+**Result Set Management CUJ** \- As a user I want to dictate which fields are returned in a search
+
+- Requirement: Support query controls to set the desired returned fields.
+
+**ACLs CUJ** \- As a user of Access Control Lists (ACLs), I want to manage user-based permissions, specific command permissions, etc to manage vector search access
+
+- Requirement: Existing ACls functionality should extend to vector search commands
```
zuiderkwast, line 79: I'm missing some details about ACL.

* ACL per command automatically extends to module commands, so we get these automatically, right?
* Does the module create a new ACL category for the VSS commands? (There's module API for that, IIRC.)
* ACL rules per key pattern, do these extend to searches? How/why not?

VSS.md
```diff
@@ -0,0 +1,518 @@
...
+- Geospatial: Allows storing, querying, and manipulating location-based data such as radius queries, nearest neighbor searches, and distance calculations, leveraging sorted sets and geohashes.
+
+**Functionality Gaps**
+- Non-vector queries: Currently, non-vector indexes are utilized only within the context of vector search queries. Expanding support to standalone non-vector use cases broadens applicability. For instance, a query could search for products priced within a specific range, tagged as "sport," and with a title containing the word "basketball." An example syntax for such a query might look like:  
+
+```
+FT.SEARCH index_name "@price:[min max] @tags:{sport} @title:basketball"
+```
+
+- ACLs: allowing query and index operations to respect the permissions configured for each user.
```
zuiderkwast, line 446: Is this about the ACL rules for key patterns?

What's the idea here? Should users be able to search only for keys they have access to read?

Technically, how could this work? Search all keys first and then filter before returning the results to the client?

VSS.md
allenss-amazon: None of the search commands actually have keys on the command, so the existing built-in ACL support for modules doesn't apply. Separately AWS has implemented module extensions for ACLs for search. This will be contributed, but under a separate RFC.

VSS.md
allenss-amazon: The security model that AWS implemented directly supports the prefix-based definition of indexes. This implementation operates at the start of a command and checks if the current user has read access to any of the keys that MIGHT be in the index, i.e., current user must have read access to 100% of the keyspace covered by the PREFIX clause of the creation of the index. Once that check passes, no additional security check would be performed.

VSS.md
zuiderkwast: ACL is not only about keys. You can allow and disallow commands:

       • +<command>: Add the command to the list of commands the user can call.  Can be used with | for allowing subcommands (e.g “+config|get”).

       • -<command>:  Remove the command to the list of commands the user can call.  Starting Valkey 7.0, it can be used with | for blocking sub‐
         commands (e.g “-config|set”).

hwware:
Can we close RFC https://github.com/valkey-io/valkey-rfc/pull/8?

VSS.md
hwware, line 96: I guess the word "distance" is related to vector, but here we're missing how distance  comes.

VSS.md
```diff
@@ -0,0 +1,518 @@
...
+## Current Specification
+
+The search module allows a user to create any number of indexes. Each index consists of any number of typed fields and covers a specified subset of the keys within a Valkey database.
+
+The creation of an index initiates a backfill process in the background that will scan the keyspace and insert all covered keys into the newly created index. Mutations of a key are automatically reflected in the index. This process is automatic after the creation of the index. Mutation of keys is explicitly allowed during the backfill process. The current state of a backfill may be interrogated by a client using the FT.INFO command.
+
+A query operation is a two step process. In the first step, a query is executed that generates a set of keys and distances. In the second step, that list of keys and distances is used to generate a response.
+
+The first step is performed in a background thread, while the second step is performed by the mainthread. If a key produced by step 1 is mutated before step 2 is executed by the mainthread, the result of the query may not be consistent with the latest state of the keys.
+
```
hwware, line 99: Can we tolerance this inconsistent?

VSS.md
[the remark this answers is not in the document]
yairgott: Right but this ones is specific to ft.search

VSS.md
yairgott: I agree that the API is not intuitive and has room for improvement. However, we've opted for RediSearch compatibility to facilitate easier adoption. In the long run, though, we should explore designing a more user-friendly API.

VSS.md
yairgott: Right, module-added commands are already supported by ACL. We will flesh out the details of how ACL will work in a separate RFC.

VSS.md
yairgott: hmm, I think that we should discuss this. I do have some reservations about this approach because ACLs might be changed after an index is created and ft.search should still honor the ACLs.

VSS.md
allenss-amazon: In CMD mode, the application could use LUA or multi/exec to avoid this issue-- with the concurrent reduction in throughput.

For CME, the question is more interesting as there is no attempt to synchronize the per-shard queries. Thus while it may be possible to avoid this issue in the originating shard (i.e., do something like the LUA multi/exec described above), the application will still have to deal with the fact that the returned results might be out-of-date for all of the non-originating shards.

VSS.md
allenss-amazon: Perhaps my explanation was poorly worded. The ACL checks operate on every command. My comment about "no additional security check would be performed" meant to say that there's no key-level access checking. An individual user either has 100% or 0% access to an index which is determined right at the start of the command.

VSS.md
```diff
@@ -0,0 +1,518 @@
...
+  - **TAG**: A tag field is a string that contains one or more tag values.  
+      
+    - **SEPARATOR \<sep\>** (optional): One of these characters `,.<>{}[]"':;!@#$%^&*()-+=~` used to delimit individual tags. If omitted the default value is `,`.  
+    - **CASESENSITIVE** (optional): If present, tag comparisons will be case sensitive. The default is that tag comparisons are NOT case sensitive
+
+  
+
+  - **NUMERIC** Field contains a number.  
+      
+  - **VECTOR**: Field contains a vector. Two vector indexing algorithms are currently supported: HNSW (Hierarchical Navigable Small World) and FLAT (brute force). Each algorithm has a set of additional attributes, some required and other optional.
```
hwware, line 161: Do we have a chance to support GRAPH_PQ ( combination of the HNSW algorithm and the PQ algorithm), IVF_GRAPH ( combination of IVF and HNSW) and IVF_GRAPH_PQ (combination of the PQ algorithm with the IVF or HNSW). These algorithms could work for huge amount records (from hundreds of millions to  more than 1 billion files in shards)

VSS.md
```diff
@@ -0,0 +1,518 @@
...
+  
+
+  - **NUMERIC** Field contains a number.  
+      
+  - **VECTOR**: Field contains a vector. Two vector indexing algorithms are currently supported: HNSW (Hierarchical Navigable Small World) and FLAT (brute force). Each algorithm has a set of additional attributes, some required and other optional.  
+      
+    - **FLAT:** The Flat algorithm provides exact answers, but has runtime proportional to the number of indexed vectors and thus may not be appropriate for large data sets.  
+      - **DIM \<number\>** (required): Specifies the number of dimensions in a vector.  
+      - **TYPE FLOAT32** (required): Data type, currently only FLOAT32 is supported.  
+      - **DISTANCE\_METRIC \[L2 | IP | COSINE\]** (required): Specifies the distance algorithm
```
hwware, line 166: Do we have a chance to support hamming: Hamming distance ?

VSS.md
```diff
@@ -0,0 +1,518 @@
...
+
+When `NOCONTENT` is specified, each entry in the response contains only the matching keyname, Otherwise, each entry includes the matching keyname, followed by an array of the returned fields.
+
+The result fields for a key consists of a set of name/value pairs. The first name/value pair is for the distance computed. The name of this pair is constructed from the vector field name prepended with "\_\_" and appended with "\_score" and the value is the computed distance. The remaining name/value pairs are the members and values of the key as controlled by the `RETURN` clause.
+
+The query string conforms to this syntax:
+
+```
+<filtering>=>[ KNN <K> @<vector_field_name> $<vector_parameter_name> <query-modifiers> ]
+```
```
hwware, line 271: Do we have a chance to implement the Boolean query, ScriptScore Query and Re-Score Query (If the GRAPH_PQ or IVF_GRAPH_PQ index is used)

hwware:
I just add comments for this RFC, I am not sure if we could implment more index algorithms and query method, but the link Using Vector Indexes for Data Search in an Elasticsearch Cluster give me more hint https://support.huaweicloud.com/intl/en-us/usermanual-css/css_01_0123.html

VSS.md
allenss-amazon: We've taken the stance of putting into the RFC the functionality that's being contributed -- which doesn't include these particular algorithms. But with that said, there's no reason that this query architecture can't easily be extended to support these algorithms in the future.

VSS.md
allenss-amazon: Same answer as above.

hwware: Thanks for your address.

VSS.md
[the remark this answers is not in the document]
madolson: that's not my point. My point is that at least in Valkey we call errors equivalent to 400s (You send a bad request) and failure equivalent to a 500 (we couldn't handle it). Are these errors or failures? Secondly, do we really need errors for each search request? Why are these requests special?

I ask because none of the other modules do this (or the core). If we are going to be a singular project, we should have some consistently between core commands and module commands.

VSS.md
madolson: Can you flesh it out here? We want a single RFC for VSS, this is our main point of alignment and it's called out in the template as a section we would like people to address.

VSS.md
allenss-amazon: Commands that manipulate the index metadata (FT.CREATE, FT.DROP, ...) are replicated like any other mutation command. The contents of the index are not replicated, each node is expected to maintain it's own indexes derived from it's own local copy of the keys. Save/Restore of the per-field indexes is optional as they can be rebuilt at load time. The current implementation elects to save/restore vector indexes because the cost of rebuilding those is so high.    

Indexes are not built at the per-slot level, but rather at the per-shard level. Thus slot migration cannot move indexes. Rather slot migration is simply treated as per-key index update operation, i.e., delete from one shard and insert into another shard -- just like standard slot migration. 

Because query operations are broadcast to all shards, individual shards will perform their portion of the query at different points in time. This creates the situation where any key that's in the process of being migrated while the shards are being interrogated could be present in 0, 1 or 2 shard results (depending on the timing). The duplication case is easily handled during the merging of the per-shard results. The 0 occurrence case presents as a degradation of recall, not a correctness issue.

VSS.md
allenss-amazon: Implementing ACL search indexes will require an extension to the module interface and that's where the separate RFC might be required. What will be proposed is to add a module interface that performs a read/write access check based on a range of keys specified by a keyname prefix. This API would be called by every search command for each of the prefix's that are part of an index's definition. Thus a user can only execute a search query command if that user has read-access to all possible keys that could be in the index. Similarly, index create/destroy operations would require write-access to all possible keys that could be in the index.

allenss-amazon: I think this is covered in the next comment. An extension to the existing module interface is required, should that be proposed here or in a separate RFC for the core?

madolson (commented):
Core team discussion: High level consensus on the design of just copying the APIs. In the future we will consider adding our own APIs with our own semantics.

1. Should be able to run on the existing LangChain.
    1. We should be compatible with Redis' langchain code, but long term we should support our own.
2. ACLs? 
   1. We should figure out what we should do before version 1.0, add it to the open discussions. 
    2. Align with scan behavior (require * permission)
    3. Prefixes
    4. Do the check on each result.
3. Release the source code on January 15th, (Yair, can I post this?)
4. CRC hash is questionable? Why not xxHash or some modern hashing algorithm? Maybe we don't really care.
5. Module APIs needed in the core. (These need to be done for 8.1)
    1. API for reducing the overhead for clusterbus messages.
    2. Need an API so that modules can share memory. Is Memory duplication a requirement for V1?
    3. Need an API for blocking a client from a keyspace notification.

PR #17, opened 2025-01-16
RFC for Valkey-Python-Integration-Test-Framework

KarthikSubbarao:
RFC for Valkey-Python-Integration-Test-Framework.

### DIFF
--- a/ValkeyLDAP.md
+++ b/ValkeyLDAP.md
@@ -1,0 +1,189 @@
+---
+RFC: 20
+Status: Accepted
+---
+
+# ValkeyLDAP Module RFC
+
+## Abstract
+
+The proposed Valkey LDAP module, named ValkeyLDAP, supports authentication to Valkey through external to Valkey managed authentication mechanisms based on the LDAP protocol. The module provides seamless integration between Valkey's authentication system and external enterprise authentication infrastructure, allowing organizations to leverage existing identity management systems. The design emphasizes security, scalability, and minimal performance impact while providing robust user authentication.
+
+## Motivation
+
+Organizations running Valkey in enterprise environments frequently need to integrate with existing identity management systems that support the LDAP protocol so that this can be centrally managed. This module addresses these challenges by:
+
+- eliminating the need for external authentication proxies
+- reducing operational complexity
+- providing native Active Directory integration
+- enabling centralized user and permission management
+- maintaining Valkey's performance characteristics
+- maintaining Valkey's existing authenthication workflow
+- supporting compliance requirements for centralized authentication
+
+The Valkey Module API already supports the offloading of user authentication to an external module, we will make use of this API to implement the LDAP authentication module.
+
+## Design considerations
+
+This module will support the two traditional approaches to authenticate users against an LDAP server:
+* The *bind* approach.
+* The *search+bind* approach.
+
+The `bind` mode can be used when the username is part of the distinguish name (DN) of user entries in the LDAP directory, while the `search+bind` mode allows for a much more flexible LDAP directory structure.
+
+### Bind Mode Authentication
+In the `bind` mode, the module will bind to the DN constructed by prepending a configurable prefix and appending a configurable suffix to the username. Typically, the prefix parameter is used to specify `CN=`, or `DOMAIN\` in an Active Directory environment. The suffix is used to specify the remaining part of the DN in a non-Active Directory environment.
+
+### Search+Bind Authentication
+In the `search+bind mode`, the module first binds to the LDAP directory with a username and password of an account that has permissions to perform search operation in the LDAP directory. If no username and password is configured for the binding phase, an anonymous bind will be attempted to the directory.
+
+After the binding phase, a search operation is performed over the subtree at a configurable base DN string, and will try to do an exact match of the username specified in the `AUTH` command against the value of a configurable entry attribute.
+
+Once the user has been found in this search, the module re-binds to the LDAP directory as this user, using the password specified in the `AUTH` command, to verify that the login is correct.
+
+This mode allows for significantly more flexibility in where the user objects are located in the directory, but will cause two additional requests to the LDAP server to be made.
+
+### Minimum Viable Product
+
+In a first stage of the module implementation, the module will support a minimal set of features and requirements to successfully handle LDAP authentication queries in a production evironment, namely:
+
+* Scalability
+  * Efficient handling of concurrent authentication requests
+  * Asynchronous communication with LDAP servers
+  * Connection pooling to reduce latency
+* Security
+  * TLS 1.2+ support
+  * Secure credential handling and storage
+* Maintainability
+  * Comprehensive error handling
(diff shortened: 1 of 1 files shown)


zuiderkwast (commented):
Can you add examples of how to write tests corresponding to those that use multiple `start_server`, including tags for skipping test cases? For example a translation of the following test from tests/integration/replication.tcl:

```TCL
start_server {tags {"repl network external:skip"}} {
    set slave [srv 0 client]
    set slave_host [srv 0 host]
    set slave_port [srv 0 port]
    set slave_log [srv 0 stdout]
    start_server {} {
        set master [srv 0 client]
        set master_host [srv 0 host]
        set master_port [srv 0 port]

        # Configure the master in order to hang waiting for the BGSAVE
        # operation, so that the slave remains in the handshake state.
        $master config set repl-diskless-sync yes
        $master config set repl-diskless-sync-delay 1000

        # Start the replication process...
        $slave slaveof $master_host $master_port

        test {Slave enters handshake} {
            wait_for_condition 50 1000 {
                [string match *handshake* [$slave role]]
            } else {
                fail "Replica does not enter handshake state"
            }
        }

        test {Slave enters wait_bgsave} {
            wait_for_condition 50 1000 {
                [string match *state=wait_bgsave* [$master info replication]]
            } else {
                fail "Replica does not enter wait_bgsave state"
            }
        }

        # Use a short replication timeout on the slave, so that if there
        # are no bugs the timeout is triggered in a reasonable amount
        # of time.
        $slave config set repl-timeout 5

        # But make the master unable to send
        # the periodic newlines to refresh the connection. The slave
        # should detect the timeout.
        $master debug sleep 10

        test {Slave is able to detect timeout during handshake} {
            wait_for_condition 50 1000 {
                [log_file_matches $slave_log "*Timeout connecting to the PRIMARY*"]
            } else {
                fail "Replica is not able to detect timeout"
            }
        }
    }
}
```

murphyjacob4:
@KarthikSubbarao what do you think about a python interface that is similar to the TCL tests? If it is too different, I would worry about the ability to translate the tests over.

I was thinking:
 * Use pytest fixtures for shared server objects across tests
 * Use pytest markers as tags
 * Use valkey-py to connect to servers
 * Use pytest-order to maintain ordering (the TCL tests heavily rely on ordering within the file)

As an example, with Viktor's test:

```
class ValkeyServer:
    def __init__(self, config):
        self.config = config
        self.host = 127.0.0.1
        self.port = find_port()

    def __enter__(self):
        print(f"Starting ValkeyServer with config: {self.config}")
        ...
        self.client = valkey.Valkey(self.host, self.port)

    def __exit__(self, exc_type, exc_val, exc_tb):
        print("Stopping ValkeyServer")
        ...
```

```
import valkey
import pytest
import re

# Logically equivalent to tags
pytestmark = [
    pytest.mark.repl,
    pytest.mark.network,
    pytest.mark.external_skip,
]

@pytest.fixture(scope="module")
def primary():
    # Configure the primary in order to hang waiting for the BGSAVE
    # operation, so that the replica remains in the handshake state.
    with ValkeyServer(config={}) as server:
        yield server

@pytest.fixture(scope="module")
def replica():
    with ValkeyServer(config={}) as server:
        yield server

@pytest.mark.order(1)
def test_replica_enters_handshake(primary, replica) {
    replica.client.replicaof(primary.host, primary.port)
    primary.client.config_set("repl-diskless-sync", "yes")
    primary.client.config_set("repl-diskless-sync-delay", "1000")
    wait_for_condition(
        condition_fn=lambda bool(re.search(".*handshake.*", replica.client.role())),
        retries=50,
        delay_ms=1000,
        on_failure=lambda pytest.fail("Replica does not enter handshake state"),
    )
}

@pytest.mark.order(2)
def test_replica_enters_wait_bgsave(primary, replica) {
    wait_for_condition(
        condition_fn=lambda bool(re.search(".*state=wait_bgsave.*", primary.client.info("replication"))),
        retries=50,
        delay_ms=1000,
        on_failure=lambda pytest.fail("Replica does not enter wait_bgsave state"),
    )

@pytest.mark.order(3)
def test_replica_is_able_to_detect_timeout_during_handshake(primary, replica):
    # Use a short replication timeout on the replica, so that if there
    # are no bugs the timeout is triggered in a reasonable amount
    # of time.
    replica.client.config_set("repl-timeout", "5")

    # But make the primary unable to send
    # the periodic newlines to refresh the connection. The replica
    # should detect the timeout.
    primary.client.execute_command("debug", "sleep", "10")

    wait_for_condition(
        condition_fn=lambda bool(re.search(".*Timeout connecting to the PRIMARY.*", replica.std_out.contents())),
        retries=50,
        delay_ms=1000,
        on_failure=lambda pytest.fail("Replica is not able to detect timeout"),
    )
```

KarthikSubbarao:
@zuiderkwast @murphyjacob4 - Thank you both for the review.

I have added 5 examples to the RFC's `Interface` section. All 5 of these examples are possible with the [valkey-test-framework](https://github.com/valkey-io/valkey-test-framework):

1. Example of each test in a test file (class) reusing the same server instance - and having serial / ordered test execution:
2. Example of each test in a test file (class) having their own server spawned/torn down - and each test in the class has custom server setup.
3. Example of each test in a test file (class) having their own server spawned/torn down - and each test in the class has the same server setup logic.
4. Example of each test in a test file (class) reusing the same replication server instances - and having serial / ordered test execution:
5. Example of each test in a test file (class) having their own replication servers spawned/torn down - and each test in the class has custom server & replication setup.



Example 1 and Example 4 are more closer to the .tcl test interface and developers can choose to follow this pattern as needed. 

However, just want to point out a benefit of spawning a server per test: It will help isolate failures (e.g. memory leaks / corruptions) more easily since each test has a fresh server is not impacted by previous tests in the file/class

madolson:
Core meeting notes:

1. Capture in the RFC what are the main features we need to support (replication, standalone, cluster, containers, external service?)
2. Concretely come up with some examples of how an existing test will looking like when we move from TCL to Python.
3. Determine how we will make sure all of the vendors of Valkey will still be able to build (what version are we supporting and what dependencies. Currently using Python 3.x)
4. Current plan forward is to continue using the Python framework for module development to help stabilize the API. JSON, Boom, LDAP, Search and timeseries will adopt it.

@KarthikSubbarao please @ the people who are in charge of the various modules to review this state. @valkey-io/contributors please also review when you have a moment to give feedback since you will be writing tests in the future.

hpatro (commented):
I would like to see the following for easier adoption to core:

* Clustering support
* Cross version testing support
* Similar TCL test helper functionalities skipping particular tests, running subset (pytest markers), running test against a already running Valkey server. https://github.com/valkey-io/valkey/blob/unstable/tests/test_helper.tcl#L570-L629
* Well documented debugging example for the framework (Each dev had their own trick in place for TCL).
* Running tests from IDE if possible (VSCode, CLion, ...)

PR #18, opened 2025-01-18
ACL RFC for Search module

allenss-amazon:
As requested, an RFC to describe the security model for the search module. 

This also proposes a new module API to make the security model enforcement efficient.

### DIFF
--- a/PrefixACL.md
+++ b/PrefixACL.md
@@ -66,14 +66,18 @@ enum KeyPrefixCheckFlags {
     VALKEYMODULE_PREFIX_CHECK_READ_ACCESS = 1,
 }
 
-int ValkeyModule_AclCheckKeyPrefixPermissions(ValkeyModuleContext *ctx, const char *prefix, size_t bytes, size_t flags)
+int ValkeyModule_AclCheckKeyPrefixPermissions(
+    ValkeyModuleContext *ctx, 
+    ValkeyModuleString **prefixes,
+    size_t prefix_count,
+    size_t flags)
 ```
 
-The ```prefix``` parameter points to the byte string to check with its length specified by the ```bytes``` parameter.
+The ```prefixes``` parameter points to an array of byte strings to check. The number of elements in the array is in the prefix_count parameters.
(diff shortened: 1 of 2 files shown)


PrefixACL.md
```diff
@@ -0,0 +1,79 @@
...
+### Module API
+
+The following module API is proposed to be implemented in the next Valkey release:
+
+```
+enum KeyPrefixCheckFlags {
+    VALKEYMODULE_PREFIX_CHECK_READ_ACCESS = 1,
+}
+
+int ValkeyModule_AclCheckKeyPrefixPermissions(ValkeyModuleContext *ctx, const char *prefix, size_t bytes, size_t flags)
```
yairgott, line 69: 1. using ValkeyModuleString* data type eliminates the need for bytes. Also, it seems more aligned with ValkeyModule_ACLCheckKeyPermissions.
2. We need to support multiple prefixes and with this interface the caller needs to loop call the this API. Consider extending the API to support variadic arguments where each argument must be a ValkeyModuleString*. .e.g.:  int ValkeyModule_AclCheckKeyPrefixPermissions(ValkeyModuleContext *ctx, size_t flags, ...)

VSS.md
```diff
@@ -0,0 +1,518 @@
+## RFC: 8
```
yairgott, line 1: Added accidentally?

yairgott (changes requested):

VSS.md
allenss-amazon: yes

PrefixACL.md
allenss-amazon: WIll fix in update.

yairgott: SG!

PrefixACL.md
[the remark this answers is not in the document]
allenss-amazon: Since this is essentially a proposal to augment the Module API, I assumed that the core team would be more interested in it. (IMO, even though the module call fallback method is fully functional, I'm skeptical that it's really a production solution due to performance concerns).

allenss-amazon:
Need to add table of commands vs other ACL categories.

PR #19, opened 2025-02-21
Update the clause regarding RDB compatibility of ValkeyJSON

joehu21:


ValkeyJSON.md
```diff
@@ -32,8 +32,8 @@ Redis Ltd.’s RedisJSON.
 
 ### RDB Compatibility
 
-To help users migrate from Redis and RedisJSON, ValkeyJSON will support RDB compatibility with RedisJSON. ValkeyJSON will 
-be able to load RDBs generated from RedisJSON. Likewise, RedisJSON will be able to load RDBs generated from ValkeyJSON.
+ValkeyJSON supports RDB compatibility with RedisJSON v1.0.8 to v2.6.12. ValkeyJSON currently has no plan to support RDB
+compatibility with RedisJSON v2.8 or later, which has a format change for Redis 7.2 and beyond that is no longer open sourced.
```
madolson, line 36: Do we understand what this compatibility split is about?

madolson: Yeah, we've discussed this. This is because they are likely running on Redis 7.4. Can you run the same test with Valkey 8.1 (or unstable) and use the rdb-version-check relaxed?

roshkhatri: That should work for Valkey 8 I think

joehu21: Retested. We are actually compatible. Updated the description.

madolson (approved):

PR #20, opened 2025-02-24
LDAP authentication module

rjd15372:
This is an RFC for a Valkey module that handles authentication against LDAP compatible identity providers, like Active Directory.

### DIFF
--- a/ValkeyLDAP.md
+++ b/ValkeyLDAP.md
@@ -1,0 +1,271 @@
+---
+RFC: (PR number)
+Status: Proposed
+---
+
+# ValkeyLDAP Module RFC
+
+## Abstract
+
+The proposed Valkey LDAP module, named ValkeyLDAP, supports authentication to Valkey through external to Valkey managed authentication mechanisms based on the LDAP protocol. The module provides seamless integration between Valkey's authentication system and external enterprise authentication infrastructure, allowing organizations to leverage existing identity management systems. The design emphasizes security, scalability, and minimal performance impact while providing robust user authentication and role-based access control.
+
+## Motivation
+
+Organizations running Valkey in enterprise environments frequently need to integrate with existing identity management systems that support the LDAP protocol so that this can be centrally managed. This module addresses these challenges by:
+
+- eliminating the need for external authentication proxies
+- reducing operational complexity
+- providing native Active Directory integration
+- enabling centralized user and permission management
+- maintaining Valkey's performance characteristics
+- maintaining Valkey's existing authenthication workflow
+- supporting compliance requirements for centralized authentication
+
+The work done in this Redis Pull Request #11659 means authentication can be handed of to a Valkey module without affecting the Valkey core authentication.
+
+## Design considerations
+
+### Performance Impact
+
+- Minimize authentication overhead through efficient caching
+- Optimize LDAP queries and connection management
+- Implement connection pooling to reduce latency
+- Balance security with performance requirements
+- Ensure minimal impact on Valkey command execution
+
+### Scalability
+
+- Support for large user bases and group hierarchies
+- Efficient handling of concurrent authentication requests
+- Horizontal scalability in clustered environments
+- Resource usage optimization for high-load scenarios
+- Graceful degradation under heavy load
+
+### Security
+
+- Zero trust security model implementation
+- Secure credential handling and storage
+- Protection against common attack vectors
+- Audit trail for authentication events
+- TLS 1.2+ required for AD communication
+- Certificate validation
+- Secure credential storage
+- Encrypted cache entries
+
+### Maintainability
+
+- Clear separation of concerns
+- Modular design for easy updates and inclusion of new protocols
+- Comprehensive logging and monitoring
+- Simple configuration management
+- Easy troubleshooting and debugging
+
+### Compatibility
+
+- Support for various Valkey deployment models
+- Backward compatibility with existing Valkey features
+- Integration with different LDAP providers
+- Support for multiple authentication methods
+- Compatibility with existing Valkey clients
+
+### High Availabilty
+
+- Multiple identity provider server support
+- Connection failover
+- Cache replication
+- Graceful degradation
+
+### Error Handling
+
+- Invalid credentials
+- Network timeouts
+- LDAP server unavailable
+- Cache corruption
+- Configuration errors
+- Group resolution failures
+
+## Specification
+
+### Authentication flow
+
+1. Client sends AUTH command
+2. Module extracts username/password
+3. Check credential cache
+4. If not cached:
+    - connect to LDAP server
+    - perform bind operation
+    - search for user group membership
+    - cache successful result
+    - set TTL
+5. Pass AUTH/NOAUTH to Valkey core
+6. Return auth response
+
+### On module load
+
+- Load configuration from Valkey config or default values
+- Initialize data structures and connection pools
+- Register command hooks with valkey core
+- Establish initial LDAP server connections
+- Load cached credentials if persistence enabled
+- Register monitoring callbacks
+
+### Commands 
+
+1 LDAP.CONFIG SET <parameter> <value>
+
+- Sets configuration parameters for the LDAP authentication module
+- Parameters include server addresses, timeouts, cache settings, etc.
+- Returns simple string reply: OK on success
+
+2 LDAP.CONFIG GET <parameter>
+
+- Gets current configuration for specified parameter
+- Returns bulk string reply: parameter value
+
+3 LDAP.RELOAD
+
+- Reloads configuration from config file
+- Reestablishes connections with LDAP servers
+- Returns simple string reply: OK on success
+
+4 LDAP.STATUS
+
+- Returns health and statistics information
+- Format: nested hash with connection states, cache stats, auth counts
+
+5 LDAP.FLUSHCACHE
+
+- Clears the credential cache
+- Options to flush all or specific users
+- Returns integer reply: number of entries flushed
+
+### Valkey Command Hooks
+
+#### Authentication Command Hook
+
+- Intercepts AUTH command
+- Processes credentials through LDAP binding
+- Maintains original Valkey auth functionality for local accounts
+
+#### ACL Command Hook
+
+- Integrates with ACL commands
+- Maps LDAP groups to Valkey ACLs
+- Preserves ACL inheritance and evaluation logic
+
+### Internal data structures
+
+#### Configuration Store
+
+- Hash table for configuration parameters
+- Persistent across restarts via RDB or config file
+- Thread-safe access with read-write locks
+
+#### Credential Cache
+
+- LRU cache with configurable TTL
+- Encrypted storage with secure key management
+- Size-limited with automatic eviction
+- Concurrent access optimized
+
+#### Group Mapping Table
+
+- Hash table mapping LDAP groups to Valkey ACLs
+- Configurable refresh interval
+- Support for nested group resolution
+
+#### Connection Pool
+
+- LDAP connection objects with state tracking
+- Priority queue based on connection health
+- Automatic connection cycling
+- Size limits and idle timeouts
+
+### Configuration
+
+#### Server Configuration
+
+- LDAP.servers: list of LDAP server addresses (comma-separated)
+- LDAP.port: LDAP port (default: 636 for LDAPS)
+- LDAP.use_ssl: whether to use LDAPS (default: true)
+- LDAP.timeout.connect: connection timeout in ms (default: 1000)
+- LDAP.timeout.search: search operation timeout in ms (default: 2000)
+- LDAP.retry.max: maximum retry attempts (default: 3)
+- LDAP.retry.delay: delay between retries in ms (default: 500)
+
+#### Authentication Configuration
+
+- LDAP.basedn: base DN for user searches
+- LDAP.binddn: service account DN for initial bind
+- LDAP.bindpw: service account password (stored securely)
+- LDAP.user_filter: LDAP filter for user searches
+- LDAP.group_filter: LDAP filter for group membership queries
+- LDAP.default_permissions: default ACL for authenticated users
+
+#### Cache Configuration
+
+- LDAP.cache.enabled: enable credential caching (default: true)
+- LDAP.cache.ttl: cache entry TTL in seconds (default: 300)
+- LDAP.cache.size: maximum cache size (default: 10000)
+- LDAP.cache.persist: persist cache across restarts (default: false)
+- LDAP.cache.encryption: encryption method for cached credentials
+
+#### Security Configuration
+
+- LDAP.tls.cert: path to client certificate
+- LDAP.tls.key: path to client key
+- LDAP.tls.ca: path to CA certificate
+- LDAP.tls.verify: verify server certificates (default: true)
+- LDAP.tls.ciphers: allowed cipher suites
+- LDAP.tls.version: minimum TLS version (default: TLS1.2)
+
+### Metrics and Monitoring
+
+#### Exported Metrics
+- Authentication attempts (success/failure)
+- Cache performance (hit ratio, size, evictions)
+- LDAP connection status (latency, errors, timeouts)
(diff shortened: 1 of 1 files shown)


dmitrypol:
I would make a case for a more generic valkey-auth module with adapters for various mechanisms such as LDAP.  This way it can be extended to support custom auth mechanisms that different users might have

ValkeyLDAP.md
```diff
@@ -0,0 +1,271 @@
...
+- Maps LDAP groups to Valkey ACLs
+- Preserves ACL inheritance and evaluation logic
+
+### Internal data structures
+
+#### Configuration Store
+
+- Hash table for configuration parameters
+- Persistent across restarts via RDB or config file
+- Thread-safe access with read-write locks
```
zuiderkwast, line 162: Why thread-safe? Why do we need threads at all?

Can't we use non-blocking IO and use Valkey's event loop? There's the event loop API to do that: `ValkeyModule_EventLoopAdd`, etc.

ValkeyLDAP.md
```diff
@@ -0,0 +1,271 @@
...
+- Hash table mapping LDAP groups to Valkey ACLs
+- Configurable refresh interval
+- Support for nested group resolution
+
+#### Connection Pool
+
+- LDAP connection objects with state tracking
+- Priority queue based on connection health
+- Automatic connection cycling
+- Size limits and idle timeouts
```
zuiderkwast, line 182: What state tracking? I'm missing something in the design discussion or background info section about this. Is there a lot of state tied to an LDAP connection so different AUTH clients can't use the same connection at the same time?

Priority queue sounds like overkill to me. Can we use a simpler approach like a simple round-robin over the available connections?

zuiderkwast (commented):
@rjd15372 Nice but bloated. :) Does it really need all this? See comments.

----

@dmitrypol

> I would make a case for a more generic valkey-auth module with adapters for various mechanisms such as LDAP. This way it can be extended to support custom auth mechanisms that different users might have

Most of the configs and connection logic defined here is specific to LDAP I think.

The Module auth API is already the generic auth mechanism, or am I missing something? Different modules can use it to implement differt kinds of auth... Which parts could be shared for two different auth backends?

ValkeyLDAP.md
```diff
@@ -0,0 +1,271 @@
+---
+RFC: (PR number)
+Status: Proposed
+---
+
+# ValkeyLDAP Module RFC
```
madolson, line 6: One thing I would ask is maybe we should have a generic "external authentication" module, that supports federated authentication with a pluggable authentication provider if you want to add a new one.

ValkeyLDAP.md
madolson: There shouldn't be any cross thread memory accesses. Inside AWS we almost always run a second thread with a separate event loop for not user data (like doing authentication requests) just so we aren't stealing resources for command processing. This is similar to how we offload all of the IO.

ValkeyLDAP.md
```diff
@@ -0,0 +1,271 @@
...
+- Integrates with ACL commands
+- Maps LDAP groups to Valkey ACLs
+- Preserves ACL inheritance and evaluation logic
+
+### Internal data structures
+
+#### Configuration Store
+
+- Hash table for configuration parameters
+- Persistent across restarts via RDB or config file
```
madolson, line 161: This is interesting, we don't really do this at all today. I think it was a big mistake in the past to store Function information in the RDB file, but I think storing in the file file seems fine.

ValkeyLDAP.md
```diff
@@ -0,0 +1,271 @@
...
+- Priority queue based on connection health
+- Automatic connection cycling
+- Size limits and idle timeouts
+
+### Configuration
+
+#### Server Configuration
+
+- LDAP.servers: list of LDAP server addresses (comma-separated)
+- LDAP.port: LDAP port (default: 636 for LDAPS)
```
madolson, line 189: Should this just be a list of LDAP servers + ports. It could also include the TLS information. Reminds me of how it was brought up in the past that having a map or nested structure would be nice for configs.

ValkeyLDAP.md
```diff
@@ -0,0 +1,271 @@
...
+- LDAP.group_filter: LDAP filter for group membership queries
+- LDAP.default_permissions: default ACL for authenticated users
+
+#### Cache Configuration
+
+- LDAP.cache.enabled: enable credential caching (default: true)
+- LDAP.cache.ttl: cache entry TTL in seconds (default: 300)
+- LDAP.cache.size: maximum cache size (default: 10000)
+- LDAP.cache.persist: persist cache across restarts (default: false)
+- LDAP.cache.encryption: encryption method for cached credentials
```
madolson, line 211: What are the options here?

ValkeyLDAP.md
```diff
@@ -0,0 +1,271 @@
...
+- LDAP.retry.delay: delay between retries in ms (default: 500)
+
+#### Authentication Configuration
+
+- LDAP.basedn: base DN for user searches
+- LDAP.binddn: service account DN for initial bind
+- LDAP.bindpw: service account password (stored securely)
+- LDAP.user_filter: LDAP filter for user searches
+- LDAP.group_filter: LDAP filter for group membership queries
+- LDAP.default_permissions: default ACL for authenticated users
```
madolson, line 203: Is this something the module should be able to control? I would assume the default permissions should be just the default user?

ValkeyLDAP.md
```diff
@@ -0,0 +1,271 @@
...
+- INFO: normal operational events
+- DEBUG: detailed debugging information
+- TRACE: full protocol-level logging for troubleshooting
+
+### Cluster mode
+
+TBD :
+- this should follow the existing authentication rules for Valkey Cluster
+- configuration should be sync-ed across all nodes
+- authentication cache should be consistent across nodes
```
madolson, line 260: In AWS we don't have a consistent authentication cache and haven't seen major issues, but it wouldn't be hard to do that with the clusterbus.

madolson:
> Which parts could be shared for two different auth backends?

Depends on the implementation. The way we built it at AWS, doing it all async was something really would require some class to just implement: "checkAuthentication()" on some password, and everything else was plumped through. At AWS we have something called IAM, I know Oracle and GCP have similar things. LDAP I think is one of the major independent ones, but there are other authentication systems as well we could integrate with.

ValkeyLDAP.md
zuiderkwast: I guess there are pros and cons. If you run a small node on a single-threaded virtual environment, it's bad to use threads. Context switches slow down a lot. Scaling horizontally with many single-threaded nodes is how you can get most out of the CPUs.

ValkeyLDAP.md
zuiderkwast: Valkey stores it in the config file when you do CONFIG REWRITE, doesn't it?

madolson: `CONFIG REWRITE` doesn't happen on restart though, you have to manually trigger it. Maybe I just was interpreting it a bit differently.

rjd15372:
@zuiderkwast @dmitrypol @madolson thanks for the great and fast feedback. I'll answer the questions in each comment and will address your suggestions in the next revisions of this document.

> I would make a case for a more generic valkey-auth module with adapters for various mechanisms such as LDAP. This way it can be extended to support custom auth mechanisms that different users might have

My opinion is the same as @zuiderkwast . The Module API is already generic and each module should be specific to each identity provider. I know that probably there might be some code that is common to all systems, but at this point of the design it's still hard to predict how much code can be shared.

> @rjd15372 Nice but bloated. :) Does it really need all this? See comments.

Probably not. The list of features was a bit exhaustive. As you said in a another comment, we should build a version 1.0 with the MVP, and then add new features, and optimizations aferwards.

ValkeyLDAP.md
```diff
@@ -0,0 +1,271 @@
...
+
+## Design considerations
+
+### Performance Impact
+
+- Minimize authentication overhead through efficient caching
+- Optimize LDAP queries and connection management
+- Implement connection pooling to reduce latency
+- Balance security with performance requirements
+- Ensure minimal impact on Valkey command execution
```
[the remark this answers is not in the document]
rjd15372: I agree with you. We should define what is the MVP of  the product, and split the list of features and optimizations in stages of development.

ValkeyLDAP.md
```diff
@@ -0,0 +1,271 @@
...
+TBD :
+- this should follow the existing authentication rules for Valkey Cluster
+- configuration should be sync-ed across all nodes
+- authentication cache should be consistent across nodes 
+
+### Replication
+
+TDB :
+- this should follow the existing authentication rules for Valkey primaries and replicas
+- authentication cache should be consistent across primary and replica
```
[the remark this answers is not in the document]
rjd15372: Right, the sync can be done by an external entity, like the sysadmin, or a k8s operator, by replicating the configuration in each node.

ValkeyLDAP.md
```diff
@@ -0,0 +1,271 @@
...
+2 LDAP.CONFIG GET <parameter>
+
+- Gets current configuration for specified parameter
+- Returns bulk string reply: parameter value
+
+3 LDAP.RELOAD
+
+- Reloads configuration from config file
+- Reestablishes connections with LDAP servers
+- Returns simple string reply: OK on success
```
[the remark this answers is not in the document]
rjd15372: That's a a good idea.

ValkeyLDAP.md
rjd15372: I'll need to understand better what you mean by "federated authentication" in this case, because the RESP protocol does not support authentication tokens, so in order for the user to get access to a Valkey node, it has always to send the password.

I would prefer to focus on LDAP only, and then with all the lessons learned think about a more general solution.

ValkeyLDAP.md
rjd15372: what do you mean?

ValkeyLDAP.md
zuiderkwast: You're right, but it's the same with other configs. If we persist the LDAP configs, I think we want to persist the other configs too at the same time, with the same mechanism. An orthogonal problem.

madolson:
> My opinion is the same as @zuiderkwast . The Module API is already generic and each module should be specific to each identity provider. I know that probably there might be some code that is common to all systems, but at this point of the design it's still hard to predict how much code can be shared.

The only thing I disagree with is "it's hard to predict", because multiple of them have already been built but they aren't open-source because they aren't open-sourcable. I mentioned independently that AWS was working with NetApp and Oracle to open source our implementation for IAM. Our thought process was that there is a lot of shared learning in building the proper interface with the core, and that's the main code that is re-usable.

rjd15372:
> The only thing I disagree with is "it's hard to predict", because multiple of them have already been built but they aren't open-source because they aren't open-sourcable. I mentioned independently that AWS was working with NetApp and Oracle to open source our implementation for IAM. Our thought process was that there is a lot of shared learning in building the proper interface with the core, and that's the main code that is re-usable.

That's fair. My comment came from my lack of experience in interacting with different authentications systems APIs.

ValkeyLDAP.md
rjd15372: Nope. I removed this in this new version. User permissions must be set a priori when creating the ACL user.

rjd15372:
@madolson @zuiderkwast I've did a major revision to the text. Removed a lot of things and the specification is now more focused.
The idea is to start small and then grow if needed.

ValkeyLDAP.md
```diff
@@ -0,0 +1,209 @@
...
+
+### Configuration
+
+The configuration options for this module will be registered using the `ValkeyModule_RegisterStringConfig` API function, which will allow the user to set and get the options using the `CONFIG SET` and `CONFIG GET` commands.
+
+The list of configuration options is the following:
+
+- General options
+  - `ldap.auth_method`: the authentication method used. Possible values `bind`, or `search`.
+  - `ldap.servers`: list of LDAP server addresses (comma-separated) where each server address has following format `ldap[s]://<hostname>[:<port>][/<dn>]`
```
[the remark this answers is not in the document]
rjd15372: Yes, you're right. A space should be used as the separator

rjd15372:
@valkey-io/core-team the LDAP module implementation in https://github.com/valkey-io/valkey-ldap is mostly done and I'm ready to release the first version of the module, but before doing that I would like to have your approval on the current user API described in this RFC, namely the command `LDAP.STATUS` and the several configuration options described in section `Configuration`.

ValkeyLDAP.md
```diff
@@ -0,0 +1,196 @@
...
+3. Establish initial LDAP server connections
+4. Register command hooks with valkey core
+
+### Commands
+
+The configuration options of this module will be set using the standard `CONFIG SET` command.
+
+The only command exported by this module is:
+
+* `LDAP.STATUS`: returns health and statistics information for each LDAP server instance.
```
madolson, line 119: I guess why not use an info field?

ValkeyLDAP.md
```diff
@@ -169,6 +145,24 @@ The list of configuration options is the following:
...
+Each dictionary field may contain the following fields:
+* `url`: the URL of the server as specified in `ldap.servers` configuration options.
+* `status`: the possible values for this field are `healthy`, or `unhealthy`; the `healthy` value denotes that the server is connectable and responds to LDAP operations, and the `unhealthy` value denotes that the server cannot process LDAP operations.
+* `ping_time`: shows the RTT for an simple LDAP operation in milliseconds; this field is only visible for `healthy` servers.
+* `error`: a string value with the cause for setting the server as `unhealthy`ç this field is only visible for `unhealthy` servers.
+
+Example of the output returned by `INFO`:
+```
+# ldap_status
+ldap_server_0:url=ldap://<hostname>,status=unhealthy,error=<some error message>
```
zuiderkwast, line 163: Is the status and error part of the URL?

In general, comma is allowed inside an URL (IIRC).

madolson: I don't think the URL needs to be "exactly" right here. If you want the actual values, it's available in the config. I would consider just putting the hostname here.

zuiderkwast: @madolson Like this?

```suggestion
ldap_server_0:host=<hostname>,status=unhealthy,error=<some error message>
```

ValkeyLDAP.md
```diff
@@ -0,0 +1,196 @@
...
+
+  "server2_hostname": {
+    "status": "unhealthy",
+    "error": "an error message"
+  }
+}
+```
+
+
+### Configuration
```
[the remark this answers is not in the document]
rjd15372: They are all modifiable at runtime. I'll mention it in the text

ValkeyLDAP.md
```diff
@@ -0,0 +1,190 @@
...
+### Scalability of authentication requests
+
+Scalability will be achieve using two stratgegies:
+
+* LDAP connection pooling and muiltiplexing
+* Parallel authentication processes with asynchronous I/O
+
+#### LDAP connection pool
+
+The module should be able to limit the number of LDAP connections it uses and should use a connection pool to avoid the overhead of opening a new connection for each authentication request.
```
zuiderkwast, line 176: Connection pool implies that it's more than one connection?

Just one connection that is being reused would also achieve the goal of limiting the number of connections. If it supports pipelining (like Valkey) we could send another requests to the same connection before the previous one returns. If it works, it could be simpler than a "real" connection pool.

rjd15372: A connection pool implies that it might be one, or more connections.

The LDAP protocol does not support pipelining.

zuiderkwast (approved):
LGTM.

ValkeyLDAP.md
```diff
@@ -0,0 +1,189 @@
+---
+RFC: 20
+Status: Proposed
```
madolson, line 3:
```suggestion
Status: Accepted
```

madolson (approved):
@valkey-io/core-team Any concerns with finalizing this and merging it?

PR #22, opened 2025-05-07
Hash Field Expiration RFC

ranshid:


HFE.md
```diff
@@ -0,0 +1,632 @@
+# Hash Items TTL Support in Valkey RFC
```
madolson, line 1: There is a format, it would be nice if you could translate it https://github.com/valkey-io/valkey-rfc/blob/main/TEMPLATE.md.

HFE.md
```diff
@@ -0,0 +1,632 @@
...
+* **Latency** - After memory efficiency considerations we will require a solution which provides low latency for hash operations. Hash objects are expected to operate in O(1) for all single access operations (get, set, delete etc...) and we will not break this promise even for items with expiry.
+* **CPU efficiency** - After latency we priorities system CPU efficiency. For example we would like to avoid high CPU utilization caused by need to perform null active expiry checks during cron runs.
+* **Copherency** -  We would like the reported system state to match the logical state as much as possible. For example the reported number of keys per DB is expected to match the number of keys which are NOT logically expired.  
+
+## Interface Requirements
+
+### New command API
+
+The proposed API is very much identical to the Redis provided API (Redis 7.4 + 8.0). This is intentionally proposed in order to avoid breaking client applications already opted to use hash items TTL.
```
zuiderkwast, line 28: I don't know if KeyDB's API is the same. If it's different, then we should consider supporting it and explain why we don't if we don't.

zuiderkwast (commented):
I didn't read all of it, but don't spend too much time on the document. I'd like to see a working implementation. :)

HFE.md
ranshid: Yeh. I drifted from it. I can try and refactor it to match the format

HFE.md
```diff
@@ -0,0 +1,632 @@
...
+
+### Tenets
+
+* **memory efficiency** - At the highest priority we target a minimal metadata overhead used in order to store and manage items TTL. While the optimal overhead to maintain item TTL is 8 bytes (could be less if we allow keeping offsets from the existing epoch diff time), we understand that maintaining active expiry logic will require use of more bytes for metadata. We will make our top priority effort to minimize this overhead.
+* **Latency** - After memory efficiency considerations we will require a solution which provides low latency for hash operations. Hash objects are expected to operate in O(1) for all single access operations (get, set, delete etc...) and we will not break this promise even for items with expiry.
+* **CPU efficiency** - After latency we priorities system CPU efficiency. For example we would like to avoid high CPU utilization caused by need to perform null active expiry checks during cron runs.
+* **Copherency** -  We would like the reported system state to match the logical state as much as possible. For example the reported number of keys per DB is expected to match the number of keys which are NOT logically expired.
```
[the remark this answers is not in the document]
ranshid: typo. sorry meant Coherency

HFE.md
ranshid: I can look into it. I made an attempt to follow the Redis API since I guess most clients would align to it.

HFE.md
ranshid: @madolson I made small refactoring to place the content better in the template format

HFE.md
```diff
@@ -0,0 +1,632 @@
+# Hash Items TTL Support in Valkey RFC
+
+
+
+## Overview
+
+The open-source Valkey community, recognized the growing need for fine-grained control over hash object metadata by introducing Hash Field Expiration (HFE), a feature that allows setting time-to-live (TTL) for individual fields within hash objects. Driven by customer requests for more precise data management and inspired by Redis Ltd.'s recent introduction, We introduce an approach focusing on memory efficiency and performance while maintaining existing user experiences. The solution involves implementing two expiration mechanisms: lazy expiration, which releases fields when accessed, and a background cron job that scans and expires volatile objects. By developing an efficient active expiry mechanism, we aim to reduce cron cycle times and minimize tail latency, ultimately providing developers with granular control over data lifecycle management. The implementation also prioritizes API compatibility with existing Redis clients, enabling seamless migration and adoption for users seeking more sophisticated hash object management capabilities.
```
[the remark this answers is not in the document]
ranshid: I provided credit for the implementations I am aware of

HFE.md
```diff
@@ -0,0 +1,643 @@
...
+}
+```
+
+     In this example, the global context is holding the accessed object reference (the database can be taken from the                   server→current_client). It also holds a flags bitmap in order to have better fine grinned control over the item access (for         example avoiding expiring elements during large scan iterations).
+
+Pros:
+     - Minimal code changes.
+     - Less error prone - since this will be applied in every hashtable access we will reduce the risk of missing item being                 accessed in some flows. 
+Cons:
+      - Additional check for `accessElement` existence in the generic hashtable implementation (we have not yet evaluated if
```
madolson, line 426: My intuition is that this will be more degrading than we expect, but it's hard to intuit this without actually seeing the code.

madolson (commented):
Glad to see us making progress on our top feature request!

HFE.md
```diff
@@ -0,0 +1,643 @@
...
+In order to unblock workloads which are utilizing this logic we would like valkey to offer a similar ability.
+
+## Design considerations 
+
+### Requirements
+
+* Support Redis compatible API to set, get and manipulate Hash field TTL.
+* Support both Lazy and active expiry mechanisms for Hash field expirations.
+* Support Replication of elements TTL as well as expired element replication.
+* Extended support for the same functionality with Sets and Sorted sets.
```
hwware, line 23: I just go through several parts quickly, and the Sets and Sorted Set are not mentioned in this RFC, can we just remove these words?  Secondly, is there any reason a List can not have the TTL for its element?

HFE.md
```diff
@@ -0,0 +1,643 @@
...
+
+## Observability (TBD)
+
+## Design Details
+
+### Known design limitations
+
+* As item expiration will produce replication content, in some cases we will avoid applying full expiration logic.
+    the following cases will avoid lazy expiring items:
+     - during HSCAN, HGETALL, Copy (when duplicating an element) and RDB/EOF loading.
```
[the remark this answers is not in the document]
ranshid: Yes. true. I mantioned at another place that the expiration context will have the same apply logic as generic key expiration (ie when expirations are paused or on replica etc...)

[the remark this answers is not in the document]
ranshid: AFAIK keys doe NOT expire keys. it will just avoid adding them in the response

HFE.md
```diff
@@ -0,0 +1,643 @@
...
+    The same `hexpire` event will be issued for all different commands which manipulate item TTL (e.g.  `HEXPIRE`, `HEXPIREAT`, `HPEXPIRE` etc...)
+    NOTE that for some cases (e.g HSETEX, there will be 2 events issued: `HSET` and `EXPIRE`)
+* A new `hpersist` event will be issued whenever an item is persisted. this can be when `HPERSIST` was issued.
+* A new `hexpired` event will be issued whenever an item is actually being expired (either actively or lazily) 
+    NOTE 1 - for the initial implementation the plan is to emit `hexpired` event for each field expiry, however it might be a valid future performance optimization to batch multiple expirations on the same key into a single event reporting.  
+    NOTE 2 - That in case the `HEXPIRE` or any of its variants where used with '0' TTL (which results in the item being deleted) there should also be at least 2 events issued:  ``hexpire`` and `hexpired` (per each of the expired elements)
+
+## Append-only file (TBD)
+
+## RDB (TBD)
```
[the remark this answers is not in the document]
ranshid: I actually already implemented the  save/load in the draft which I plan to publish. will place the design ASAP

HFE.md
```diff
@@ -0,0 +1,643 @@
...
+    For example, random key selection is scanning random hashtable elements which is an algorithm implemented in the hashtable infrastructure. It would be challenging in case we would like to apply the lazy expiration logic on scanned elements.
+    
+    Pros:
+     - Follows the same approach as generic keys are handled
+     - Does not require introducing complicated new infrastructure.
+    Cons:
+     - Error prone - there are many cases where an item is accessed  
+     - Might require extensive code changes.
+     - In some cases can lead to performance degradation on the good-path - It is possible that in order to avoid code complexity we would consider to apply the `itemExpireIfNeeded` logic by first searching the item (provided in the command arguments) and then proceed with the normal implementation. Since double searching the element would probably maintain cache locality for the second search, in reality we observed 2-3% degradation by applying double search on items in the command processing.
```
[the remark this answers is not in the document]
ranshid: It is possible, however I wanted to check if we can have a change with minimal footprint, without sacrifice other things.

HFE.md
ranshid: Yeh. the draft will be issues this week (if all goes well)

HFE.md
ranshid: SETs and SortedSets are something we should plan for IMO. I am just not sure if we will be able to make it for 9.0 timeline with both of them so we prioritised Hashes first. SETs are the next target.

HFE.md
```diff
@@ -0,0 +1,643 @@
...
+
+* Support Redis compatible API to set, get and manipulate Hash field TTL.
+* Support both Lazy and active expiry mechanisms for Hash field expirations.
+* Support Replication of elements TTL as well as expired element replication.
+* Extended support for the same functionality with Sets and Sorted sets.
+
+### Tenets
+
+* **memory efficiency** - At the highest priority we target a minimal metadata overhead used in order to store and manage items TTL. While the optimal overhead to maintain item TTL is 8 bytes (could be less if we allow keeping offsets from the existing epoch diff time), we understand that maintaining active expiry logic will require use of more bytes for metadata. We will make our top priority effort to minimize this overhead.
+* **Latency** - After memory efficiency considerations we will require a solution which provides low latency for hash operations. Hash objects are expected to operate in O(1) for all single access operations (get, set, delete etc...) and we will not break this promise even for items with expiry.
```
xbasel, line 28: "provides low latency" is vague. Do you mean 'doesn’t regress existing latency by a noticeable margin'?

HFE.md
```diff
@@ -0,0 +1,643 @@
...
+    Second** all related item tracking can be accessed and updated from the object itself.
+
+### Lazy expiration
+
+Lazy expiration happen whenever an item which has expired TTL is accessed. In this case the item will be **immediately** removed from the owning object and expired (as described in the expiration context section). This include both read and write commands access context. For items (fields) ‘access’ logic we aim to follow the same logic used for generic keys. 
+When accessing an item, there are 3 potential states it can be stated:
+
+1. VALID - meaning the item is to be treated as any existing item and be included in replies as well as operated on during actions and mutations.     
+2. INVALID - meaning the item should be treated as “not exists” and is NOT to be included in any operations and/or replies.
+3. INVALIDATED(deleted) - meaning the item is to-be-removed immediately (i.e expired) and thus does not exist anymore.
```
xbasel, line 319: The terms **VALID**, **INVALID**, and **INVALIDATED** are too similar and not intuitive. INVALID suggests corruption, but an expired field is still well-formed and valid. Consider clearer terms like LIVE/NORMAL, EXPIRED, and DELETED/PURGED to reflect TTL state accurately.

zuiderkwast: I agree. In general, I think we should use the word "expired" to mean that the time has passed for a key, whether it still exists in memory or not. An expired key can exist in the database if we haven't deleted it yet.

We should not use this word as a transitive verb as in "to expire a key". Instead we should say "to delete an expired key".

HFE.md
```diff
@@ -0,0 +1,632 @@
...
+
+
+### Keyspace-notification requirements
+
+* A new `hexpire` event will be issued whenever a key field is being assigned with an expiry. This will include both cases of assigning TTL for element which had NO prior TTL assigned and when a TTL is changed on an item which already had TTL assigned. 
+    The same `hexpire` event will be issued for all different commands which manipulate item TTL (e.g.  `HEXPIRE`, `HEXPIREAT`, `HPEXPIRE` etc...)
+    NOTE that for some cases (e.g HSETEX, there will be 2 events issued: `HSET` and `EXPIRE`)
+* A new `hpersist` event will be issued whenever an item is persisted. this can be when `HPERSIST` was issued.
+* A new `hexpired` event will be issued whenever an item is actually being expired (either actively or lazily) 
+    NOTE 1 - for the initial implementation the plan is to emit `hexpired` event for each field expiry, however it might be a valid future performance optimization to batch multiple expirations on the same key into a single event reporting.
```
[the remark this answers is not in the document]
ranshid: I am not sure which is better. the event has no indication about the field being expired, BUT if we batch to the same event the application will not get events in the same number as the expired items. not sure which is better performance vs functionality. It is true that batching it later will probably be a breaking change (but in case it will break it also means that some applications are logically dependent on this separation :) )

HFE.md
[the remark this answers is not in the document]
ranshid: @madolson added sections for RDB and AOF

HFE.md
ranshid: IMO I was clear about asymptotic guarantees.

HFE.md
ranshid: I agree about this terminology in the scope of lazy expiration (i.e. in the RFC). I used these terms since these are the states recognized by the hashtable which I would like to keep out of scope of  "expiration" and "volatile".
I will change it in the RFC.

ranshid: @xbasel / @zuiderkwast  fixed. looks more to your point?

HFE.md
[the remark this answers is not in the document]
ranshid: regarding dynamically change the hashtable type... it is doable, small change probably, lets evaluate during the PR review?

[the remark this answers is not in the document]
ranshid: >that also includes the number of expirable items. That ways we can efficiently skip any logic if we are operating on the special type.

That one is planned and will be part of the initial draft.
we will have both `hashTypeHasVolatileItems()` and `hashTypeNumVolatileItems()`

HFE.md
```diff
@@ -415,9 +415,23 @@ in caseses were most items are not volatile and the metadata size is small (smal
...
+
+```
+db0:keys=1,expires=0,avg_ttl=0,volatile_items=16
+```
+
+NOTE - we can also consider adding keys_with_volatile_items statistic to track how many objects have 
+volatile items. eg:
+
+```
+db0:keys=1,expires=0,avg_ttl=0,volatile_items=16,keys_with_volatile_items=1
```
zuiderkwast, line 436: We don't have anything for items currently so it seems fine to me to skip the info about volatile items initially.

If we want it, then I guess we should also add the total number of items (including non-volatile) and the number of keys with items (including non-volatile items).

Regarding the nameing, shouldn't it match the naming used for keys? Here "expires" means the number of volatile keys. To match that, we could use "items", "item_expires" and "item_avg_ttl".

Later, if we add expiration on set elements, sorted set elements, etc. then all these are considered items too, right? Isn't it more useful to have the number of hash fields with expire, set elements with expire and sorted set elements with expire as separate metrics? The full picture starts to look like Wen's KEYSIZES fields: https://github.com/valkey-io/valkey/pull/1967

ranshid: >The full picture starts to look like Wen's KEYSIZES fields: https://github.com/valkey-io/valkey/pull/1967

Exactly. I think we will be able to add items statistics and such later on as part of the new KEYS observability

HFE.md
```diff
@@ -0,0 +1,847 @@
...
+* **Latency** - After memory efficiency considerations we will require a solution which provides low latency for hash operations. Hash objects are expected to operate in O(1) for all single access operations (get, set, delete etc...) and we will not break this promise even for items with expiry.
+* **CPU efficiency** - After latency we priorities system CPU efficiency. For example we would like to avoid high CPU utilization caused by need to perform null active expiry checks during cron runs.
+* **Compatability** - We will avoid breaking clients which are already using HFE API provided by other providers.
+* **Coherency** -  We would like the reported system state to match the logical state as much as possible. For example the reported number of keys per DB is expected to match the number of keys which are NOT logically expired.
```
JimB123, line 36: I'm not completely sure what you're saying here.  Are you comparing items in a hash to items in the DB?

From the DB perspective, when we perform INFO, the number of keys reported in the last line will include keys which have passed the expiration time but have not yet been physically deleted, right?

Are you suggesting that HASHes should behave "properly"?  and only report the number of unexpired items?

HFE.md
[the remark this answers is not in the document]
JimB123: Are there any other examples of keyspace events which mention multiple keys(fields)?  For instance, does HMSET, generate a single event?  or multiple?

If there are no examples of events spanning multiple keys/fields, I think it's better to maintain consistency & simplicity.  It's very probable that nobody will subscribe to these events, which make performance a non-issue.

HFE.md
```diff
@@ -0,0 +1,847 @@
...
+We suggest to avoid placing new configuration to fine tune this feature. Internal configurations can be discussed in order to be able to
+prevent active expiration of volatile items.
+
+## Observability
+
+AT the first drop we will only introduce some basic statistics:
+1. volatile_items will be added to the Keyspace section per-db line. eg:
+
+```
+db0:keys=1,expires=0,avg_ttl=0,volatile_items=16
```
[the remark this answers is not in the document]
JimB123: I recommend against modification of this existing INFO line.
This line - containing the number of keys - is one of the most likely lines in INFO to be parsed by a client application (or even a client library).  Altering this line has the opportunity to break a large number of existing client applications.

I suggest you add an additional line (or lines) containing the new information.

HFE.md
JimB123: Again, I suggest a new line.  Also, in the new line, you can't use a prefix like `db0:` as this might be searched for explicitly.

HFE.md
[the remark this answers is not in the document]
JimB123: Said another way...  You're saying that READ commands (like HGETALL) shouldn't perform modifications (WRITES) to the data structure.  Right?

Modification of the structure (even deleting logically expired data) should be avoided during a "READ" operation.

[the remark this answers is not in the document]
JimB123: Yes, we don't want READ commands to be generating replication traffic.

HFE.md
zuiderkwast: I just checked what redis does. From their docs of the INFO command, we can see they added "subexpiry":

```
# Keyspace

db0:keys=112125,expires=456,avg_ttl=31368299122246,subexpiry=0
```

HFE.md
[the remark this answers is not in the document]
zuiderkwast: We can hope that clients that parse this split by comma and can handle extra fields. It may break a few but the alternative (a new line per db) may be uglier.

Redis added "subexpiry=0". Can we do the same? We copy their command API...

HFE.md
```diff
@@ -0,0 +1,847 @@
...
+`HPEXPIRETIME` has the same semantics as `HEXPIRETIME`, but returns the absolute Unix expiration timestamp in milliseconds since Unix epoch instead of seconds.
+
+
+## Keyspace-notification requirements
+
+* A new `hexpire` event will be issued whenever a key field is being assigned with an expiry. This will include both cases of assigning TTL for element which had NO prior TTL assigned and when a TTL is changed on an item which already had TTL assigned. 
+    The same `hexpire` event will be issued for all different commands which manipulate item TTL (e.g.  `HEXPIRE`, `HEXPIREAT`, `HPEXPIRE` etc...)
+    NOTE that for some cases (e.g HSETEX, there will be 2 events issued: `HSET` and `HEXPIRE`)
+* A new `hpersist` event will be issued whenever an item is persisted. this can be when `HPERSIST` was issued.
+* A new `hexpired` event will be issued whenever an item is actually being expired (either actively or lazily)
```
zuiderkwast, line 228: Regarding "being expired" terminology, I think the word expired should always refer to logically expired.

```suggestion
* A new `hexpired` event will be issued whenever an expired item is detected and deleted (either actively or lazily) 
```

HFE.md
ranshid: >Are you suggesting that HASHes should behave "properly"? and only report the number of unexpired items?

I was mainly setting a tenet here. We will not fully support this at this point (for example HLEN will, return the number of items even though some of the items have already been expired). In some implementations, though we can basically know EXACTLY how many items are expired. For example in case we track all volatile hash items in rax 9like the client's timeout rax) we can provide an O(1) report for the number of items which are already expired. However this would cost much memory to maintain, thus this is currently avoided and the fact that this tenet is lower priority than the memory efficiency is providing the justification for such a decision.

HFE.md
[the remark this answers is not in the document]
ranshid: >Are there any other examples of keyspace events which mention multiple keys(fields)? For instance, does HMSET, generate a single event? or multiple?

Most commands generate a single event per command and not per field. For example hdel is reported once per command. In this proposal the hexpire event will ALSO be reported once per command and not per item. the porposal DOES state that the new `hexpired` event will be reported once per item. However I can think of improving that if we decide to. For example we can identify that we expired items in the start/end of the access context and decide to issue the event in the closing of the access context.

HFE.md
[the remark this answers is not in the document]
ranshid: >Modification of the structure (even deleting logically expired data) should be avoided during a "READ" operation.

No. saying that would basically mean we do not support lazy expirations. I am saying that there is a common expiration logic which is applied for generic keys. currently this logic can be observed by reading the content of `expireIfNeededWithDictIndex` which performs various checks on importing mode, replication client, eviction pause etc... we will also comply with the same logic when deciding to expire hash items.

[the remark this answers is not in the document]
ranshid: >Yes, we don't want READ commands to be generating replication traffic.

Well, we probably don't want it because of the implementation complexity, however lazy expiring keys is basically what this feature provides. we can decide to avoid READ commands lazy expirations if we want to.

HFE.md
```diff
@@ -0,0 +1,847 @@
...
+
+#### How do we limit the number of objects?
+
+As mentioned before, we cannot allow logarithmic complexity. this means we will have to keep the list small enough (say 512 elements). But what happens when we have more than 512 elements with expiry data?
+For that we will manage The RAX data structure mentioned before. Whenever a skip list has grown over the maximum amount of Objects we will split it to 2 lists. SkipLists are great in supporting range split and merges. We can locate the median value in a logarithmic time and since the list is already sorted.
+
+### Active expiry cycle and credits 
+
+We plan to introduce a new type of active expiry cycle (in addition to `ACTIVE_EXPIRE_CYCLE_FAST` and `ACTIVE_EXPIRE_CYCLE_SLOW` ): `ACTIVE_EXPIRE_CYCLE_ITEMS`.
+The new expiry cycle will use the same overall logic as the regular active expiry cycle with the following adjustments:
```
xbasel, line 786: Key active expiry is built around iterating over databases, and for each DB it scans db->expires using a hit-or-miss approach over keys with expiry. Hash field expiry doesn’t follow the same model, it’s not based on random sampling, and its pacing is entirely different.

I think trying to couple field expiry with key expiry would require a significant refactor, and I don’t see a clear benefit. Expiring keys and fields are logically unrelated and should remain decoupled. I don’t see how field expiry fits naturally into the existing loops in `activeExpireCycle()`. Entry expiry should have its own database iterator IMO.

Using the same logic for both means combining two almost entirely independent mechanisms into one, which adds complexity without meaningful gain. Is it really worth it?

zuiderkwast: Separate sounds good to me, unless Ran has some good motivation for coupling them.

HFE.md
```diff
@@ -0,0 +1,847 @@
...
+
+**Synopsis** 
+
+```
+HPERSIST key FIELDS numfields field [field ...]
+```
+
+Remove the existing expiration on a hash key's field(s), turning the field(s) from *volatile* (a field with expiration set) to *persistent* (a field that will never expire as no TTL (time to live) is associated).
+
+#### HSETEX
```
yangbodong22011, line 169: Why we have two `HSETEX` command?  same with L44?

HFE.md
```diff
@@ -0,0 +1,847 @@
...
+#### HPEXPIRETIME
+
+**Synopsis**
+
+```
+HPEXPIRETIME key FIELDS numfields field [field ...]
+```
+
+`HPEXPIRETIME` has the same semantics as `HEXPIRETIME`, but returns the absolute Unix expiration timestamp in milliseconds since Unix epoch instead of seconds.
+
```
yangbodong22011, line 220: We don't support `HGETDEL` command?

ranshid:
Note that this is already completed via: https://github.com/valkey-io/valkey/pull/2089

HFE.md
```diff
@@ -0,0 +1,847 @@
+---
+RFC: 22
+Status: Proposed
```
zuiderkwast, line 3:
```suggestion
Status: Approved
```

Implicitly approved by merging and releasing the feature. :)

HFE.md
```diff
@@ -0,0 +1,847 @@
...
+* `EX seconds` — Set the specified expiration time in seconds.
+* `PX milliseconds` — Set the specified expiration time in milliseconds.
+* `EXAT unix-time-seconds` — Set the specified Unix time in seconds at which the fields will expire.
+* `PXAT unix-time-milliseconds` — Set the specified Unix time in milliseconds at which the fields will expire.
+* `KEEPTTL` — Retain the TTL associated with the fields.
+
+The `EX`, `PX`, `EXAT`, `PXAT`, and `KEEPTTL` options are mutually exclusive.
+
+
+#### HEGTEX
```
zuiderkwast, line 69:
```suggestion
#### HGETEX 
```

Typo

HFE.md
```diff
@@ -0,0 +1,847 @@
...
+
+1. We can bound the number of low resolution buckets so that even if we have less elements per bucket they will not live for a long time (the total range of the higher resolution layer)
+2. Higher we extend the buckets resolution, we increase the probability of placing many items in the same bucket. 
+3. When scanning items in the highest resolution layer, we only iterate over items which are expired. 
+
+The main issue with timer wheels is the cascading operation which might be costly since it will require to iterate over all the items in the cascaded bucket. It is important to note that the higher the level resolution gets, the less frequent this cascading operation needs to take place and it can also increase the probability that these items will be lazily deleted reducing the need to actively iterate and expire them.
+I will not get into the details but there are many ways we can optimize the cascading process. For example we can improve memory access by batch prefetching the items. We can also track the size, mean and variance in each bucket and take cascading decisions based on them.
```
zuiderkwast, line 847: Add in the end of the document:

> ## Links
>
> * Original issue
> * PR
> * Blog post

zuiderkwast (commented):
Let's not spend too much time but add links and maybe mention in which version it was released.

HFE.md
zuiderkwast: Yeah, it's duplicated. Let's delete one of them.

PR #24, opened 2025-07-14
Valkey FullTextSearch RFC

KarthikSubbarao:
Valkey FullTextSearch RFC

Content taken from: https://github.com/valkey-io/valkey-search/tree/fulltext/rfc

I can help update the RFC based on any question / comment added to the PR

### DIFF
--- a/ValkeyFullTextSearch.md
+++ b/ValkeyFullTextSearch.md
@@ -1,0 +1,392 @@
+---
+RFC: (PR number)
+Status: (Change to Proposed when it's ready for review)
+---
+
+# Title
+Text Searching
+
+## Abstract
+
+The existing search platform is extended with a new field type, ```TEXT``` is defined. The query language is extended to support locating keys with text fields based on term, wildcard, fuzzy and exact phrase matching.
+
+## Motivation
+
+Text searching is useful in many applications.
+
+## Terminology
+
+In the context of this RFC.
+
+| Term        | Meaning                                                                                                                                                                                                                     |
+| ----------- | --------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------- |
+| key         | A Valkey key. Depending on the context of usage could be the text of the key name or the contents of the HASH or JSON key.                                                                                                  |
+| field       | A component of an index. Each field has a type and a path.                                                                                                                                                                  |
+| index       | A collection of fields and field-indexes. The object created by the ```FT.CREATE``` command.                                                                                                                                |
+| field-index | A data structure associated with a field that is intended to accelerate the operation of the search operators for this field type.                                                                                          |
+| character   | An Unicode character. A character may occupy 1-4 bytes of a UTF-8 encoded string.                                                                                                                                           |
+| word        | A syntactic unit of text consisting of a vector of characters. A word is delimited by un-escaped punctuation and/or whitespace.                                                                                             |
+| token       | same as a word                                                                                                                                                                                                              |
+| text        | A UTF-8 encoded string of bytes.                                                                                                                                                                                            |
+| stemming    | A process of mapping similar words into a common base word. For example, the words _drives_, _drove_ and _driven_ would be replaced with the  _drive_.                                                                      |
+| term        | The output of stemming a word.                                                                                                                                                                                              |
+| prefix tree | A mapping from input string to an output object. Insert/delete operations are O(length(input)). Additional operations include the ability to iterate over the entries that share a common prefix in O(length(prefix)) time. |
+
+## Design considerations
+
+The text searching facility provides machinery that decomposes text fields into terms and field-indexes them.
+The query facility of the ```FT.SEARCH``` and ```FT.AGGREGATE``` commands is enhanced to select keys based on combinations of term, fuzzy and phrase matching together with the existing selection operators of TAG, Numeric range and Vector searches.
+
+### Tokenization process
+
+A tokenization process is applied to strings of text to produce a vector of terms.
+Tokensization is applied in two contexts. 
+First as part of the ingestion process for text fields of keys.
+Second to process query string words and phrases.
+
+The tokenization process has four steps.
+
+1. The text is tokenized, removing punctuation and redundant whitespace, generating a vector of words.
+2. Latin upper-case characters are converted to lower-case.
+3. Stop words are removed from the vector.
+4. Words with more than a fixed number of characters are replaced with their stemmed equivalent term according to the selected language.
+```
+┌─────────────────┐     ┌───────────────┐     ┌──────────────┐     ┌───────────────┐
+│  Input Text     │     │ Tokenization  │     │ Case         │     │ Stop Word     │
+│  "The Running   │ ──> │ ["The",       │ ──> │ ["the",      │ ──> │ ["running",   │
+│   Searches cat" │     │  "Running",   │     │  "running",  │     │  "searches",  │
+└─────────────────┘     │  "Searches",  │     │  "searches", │     │  "cat"]       │
+                        │  "cat"]       │     │  "cat"]      │     └───────┬───────┘
+                        └───────────────┘     └──────────────┘             │
+                                                                           V
+                                                                      ┌───────────────┐
+                                                                      │ Stemming      │
+                                                                      │ ["run",       │
+                                                                      │  "search",    │
+                                                                      │  "cat"]       │
+                                                                      └───────────────┘
+```
+
+Step 1 of the tokenization process is controlled by a specified list of punctuation characters. 
+Sequences of one or more punctuations characters delimit word boundaries.
+Note that individual punctuation characters can be escaped using the normal backslash prefix syntax, causing the next character to be treated as non-punctuation.
+
+### Text Field-Index Structure
+
+The text field-index is a mapping from a term to a list of locations.
+Depending on the configuration, the list of locations may be just a list of keys or a list of key/term-offset pairs. The list of locations is also known as a postings or a postings list.
+
+The mapping structure is built from one or two prefix trees. When present, the second prefix tree is built from each inserted term sequenced in reverse order, effectively providing a suffix tree.
+
+Here is a visual example. Note that in the drawing, the sections noted as "same postings as _xxxx_" mean that the pointer actually points to the same memory area as referenced. In other words, when a suffix tree is present, the postings list is referenced by both (think shared_ptr).
+
+```
+┌──────────────────────────────────────────────────────────────────────────────┐
+│                                  TextIndex                                   │
+├──────────────────────────────────────────────────────────────────────────────┤
+│                                                                              │
+│ ┌──────────────────┐                                ┌──────────────────┐     │
+│ │   Prefix Tree    │             Postings           │   Suffix Tree    │     │
+│ └──────────────────┘                                └──────────────────┘     │
+│ │                  │                                │                  │     │
+│ │  ┌────────────┐  │          ┌────────────┐        │  ┌────────────┐  │     │
+│ │  │   apple    │──┼─────────>│ key1: [0]  │<───────┼──│   elppa    │  │     │
+│ │  └────────────┘  │          │ key2: [3]  │        │  └────────────┘  │     │
+│ │                  │          └────────────┘        │                  │     │
+│ │                  │                                │                  │     │
+│ │  ┌────────────┐  │          ┌────────────┐        │  ┌────────────┐  │     │
+│ │  │   banana   │──┼─────────>│ key1: [1]  │<───────┼──│   ananab   │  │     │
+│ │  └────────────┘  │          │ key3: [2]  │        │  └────────────┘  │     │
+│ │                  │          └────────────┘        │                  │     │
+│ │                  │                                │                  │     │
+│ │  ┌────────────┐  │          ┌────────────┐        │  ┌────────────┐  │     │
+│ │  │   cherry   │──┼─────────>│ key2: [0]  │<───────┼──│   yrrehc   │  │     │
+│ │  └────────────┘  │          │ key3: [8]  │        │  └────────────┘  │     │
+│ │                  │          └────────────┘        │                  │     │
+│ │                  │                                │                  │     │
+│ └──────────────────┘                                └──────────────────┘     │
+│                                                                              │
+└──────────────────────────────────────────────────────────────────────────────┘
+```
+
+
+### Search query operators
+
+Unlike the Vector, Tag and Numeric search operators the specification of a field is optional. If the field specification is omitted then a match is declared if any text field matches the search operator.
+
+#### Term matching
+
+There are three types of term matching: exact, wildcard and fuzzy. Exact term matching is self-descriptive, i.e., only keys containing exactly the specified text are matched.
+
+Wildcard matching provides a subset of reg-ex style matching of terms.
+Initially, only a single wildcard specifier ```*``` is allowed which matches any number of characters in a term.
+The wildcard can be at any position within the term, providing prefix, suffix and infix style matching.
+Note, in this proposal, the second prefix tree is required to perform infix and suffix matching.
+
+#### Fuzzy matching
+
+Fuzzy word matching is specified by enclosing a word in pairs of percent signs ```%```. This term matches words within one [Levenshtein edit distance](https://en.wikipedia.org/wiki/Levenshtein_distance) edit distance for each pair of percent signs from the specified word. 
+
+#### Phrase matching
+
+Exact phrase matching is specified by enclosing a sequence of term-specifiers in double quotes ```"```. In order to perform phrase matching, the field-index must be configured to contain offsets. In the context of phrase matching a term-specifier is either a term match or a fuzzy match specifier.
+
+The syntax of the phrase matching term allows the specification of two additional modifiers: `Inorder` and `Slop`. The `Slop` modifier controls the maximum distance between the matching terms. A `Slop` value of zero means that the matched terms must be sequentially in order. A `Slop` value of 1 indicates that the matched terms could have up to 1 addition non-matched word between them. The default value for `Slop` is zero.
+
+The `Inorder` modifier is a boolean which controls whether the ordering of the matches terms must match the ordering in the phrase. If `Inorder` is `false` then the terms can be matched in any order. The default is `true`.
+
+### Example Search operations
+
+These examples provide an example query string as a block of text. When used in an FT.SEARCH or FT.AGGREGATE command, this text is the contents of a single field in that command.
+Thus if the command is submitted through some application like valkey-cli which will subject its command line to various shell-like metadata operations (quote removal, environment variable substitution, etc.) then the reader must compensate for that, for example by escaping spaces which valkey-cli would tend to separate into separate command fields.
+
+#### Term Matching Examples
+
+| Example | Interpretation | Required Index Options |
+|:-:|---|:-:|
+| `abcd` | exact term match | none |
+| `a*`   | matches any term starting with `a` | none |
+| `fred*` | matches any term starting with `fred` | none |
+| `*z` | matches any term ending with `z` | SUFFIXTRIE |
+| `a*b`  | matches any term starting with `a` and ending with `b` | SUFFIXTRIE |
+
+#### Fuzzy Matching Examples
+
+| Example | Interpretation |
+|:-:|---|
+| %abc% | Matches any term within one edit-distance of `abc` |
+| %%abc%% | Matches any term within two edit-distances of `abc` |
+
+#### Phrase Matching Examples
+
+All phrase matching queries require the `WITHOFFSETS` index creation option.
+
+| Example | Interpretation |
+|:-:|---|
+| `"a b"` | Matches term `a` immediately followed by term `b` |
+| `"a %b%"` | Matches term `a` immediately followed by any term within one edit-distance of `b` |
+| `"a*b c"` | Matches any term that starts with `a` and ends with `b` followed by the term `c` |
+| `"a b c"` | Matches the terms `a`, `b` and `c` in that order with no intervening words |
+| `"a b c" => [$inorder:false]` | Matches the terms `a`, `b`, and `c` in any order with no intervening words |
+| `"a b" => [$inorder:false, $slop:1]"` | Matches the terms `a` and `b` in any order with up to one intervening term |
+
+### Search Query Operations Flow
+
+The following diagram illustrates the flow of a text search operation from query to results:
+
+```
+┌──────────────────┐
+│ Search Query     │
+│ e.g., "quick     │
+│ brown fox"       │
+└────────┬─────────┘
+         │
+         ▼
+┌──────────────────┐     ┌──────────────────────────────────────────┐
+│ Query Parser     │────▶│ Parsed Query Tree                        │
+└────────┬─────────┘     │                                          │
+         │               │    ┌─────────┐                           │
+         │               │    │ PHRASE  │                           │
+         │               │    └────┬────┘                           │
+         │               │         │                                │
+         │               │  ┌──────┴──────┼─────────────┐           │
+         │               │  │             │             |           │
+         │               │  ▼             ▼             ▼           │
+         │               │ ┌─────┐     ┌─────┐     ┌─────┐          │
+         │               │ │quick│     │brown│     │ fox │          │
+         │               │ └─────┘     └─────┘     └─────┘          │
+         │               └──────────────────────────────────────────┘
+         ▼
+┌──────────────────┐
+│ Query Execution  │
+└────────┬─────────┘
+         │
+         ▼
+┌──────────────────────────────────────────────────────────────────────┐
+│ Term Dictionary Lookup                                               │
+│                                                                      │
+│ ┌─────────────┐         ┌─────────────┐         ┌─────────────┐      │
+│ │Term: "quick"│         │Term: "brown"│         │Term: "fox"  │      │
+│ └──────┬──────┘         └──────┬──────┘         └──────┬──────┘      │
+│        │                       │                       │             │
+│        ▼                       ▼                       ▼             │
+│ ┌──────────────┐        ┌──────────────┐        ┌──────────────┐     │
+│ │Postings List │        │Postings List │        │Postings List │     │
+│ └──────┬───────┘        └──────┬───────┘        └──────┬───────┘     │
+│        │                       │                       │             │
+└────────┼───────────────────────┼───────────────────────┼─────────────┘
+         │                       │                       │
+         └───────────────────────┼───────────────────────┘
+                                 │
+                                 ▼
+┌───────────────────────────────────────────────────────────────────┐
+│ Position Filtering (for phrase matching)                          │
+│                                                                   │
+│ Check if terms appear in correct sequence with proper spacing     │
+│                                                                   │
+│ ┌─────────────────────────────────────────────────────────────┐   │
+│ │ Document: key1                                              │   │
+│ │                                                             │   │
+│ │ [quick:pos2]───┐                                            │   │
+│ │                │                                            │   │
+│ │ [brown:pos3]───┼─── Sequential? YES                         │   │
+│ │                │                                            │   │
+│ │ [fox:pos4]─────┘                                            │   │
+│ └─────────────────────────────────────────────────────────────┘   │
+│                                                                   │
+└───────────────────────────────────────────────────────┬───────────┘
+                                                        │
+                                                        ▼
+                                             ┌───────────────────┐
+                                             │ Results           │
+                                             │ [key1, key5, ...] │
+                                             └───────────────────┘
+```
+
+## System Architecture
+
+The following diagram illustrates the overall architecture and data flow of the text search system:
+
+```
+┌───────────────────────────────────────────────────────────────────────────────┐
+│                        Text Search System Architecture                        │
+├───────────────────────────────────────────────────────────────────────────────┤
+│                                                                               │
+│  ┌───────────────┐        ┌────────────────┐         ┌────────────────────┐   │
+│  │ Client        │        │ Command        │         │ Search Query       │   │
+│  │ Applications  │───────▶│ Processing     │────────▶│ Execution Engine   │   │
+│  └───────────────┘        │ (FT.SEARCH)    │         └──────────┬─────────┘   │
+│                           └────────────────┘                     │            │
+│                                    ▲                             │            │
+│                                    │                             │            │
+│                                    │                             ▼            │
+│                                    │                  ┌────────────────────┐  │
+│                                    │                  │ Text Index         │  │
+│                                    │                  │                    │  │
+│                                    │                  │ - Prefix Tree      │  │
+│                                    │                  │ - Suffix Tree      │  │
+│                                    │                  │ - Postings Lists   │  │
+│                                    │                  └──────────┬─────────┘  │
+│                                    │                             │            │
+│                           ┌────────┴───────┐                     │            │
+│  ┌───────────────┐        │ Result         │◀────────────────────┘            │
+│  │ Client        │◀───────│ Formatting     │                                  │
+│  │ Applications  │        │                │                                  │
+│  └───────────────┘        └────────────────┘                                  │
+│                                                                               │
+│                                                                               │
+│  ┌───────────────┐        ┌────────────────┐         ┌────────────────────┐   │
+│  │ Client        │        │ Command        │         │ Text Index         │   │
+│  │ Applications  │───────▶│ Processing     │────────▶│ Builder            │   │
+│  └───────────────┘        │ (FT.CREATE)    │         └──────────┬─────────┘   │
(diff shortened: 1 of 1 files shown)


ValkeyFullTextSearch.md
```diff
@@ -0,0 +1,392 @@
+---
+RFC: 24
+Status: (Change to Proposed when it's ready for review)
```
madolson: Is it ready for review?

ValkeyFullTextSearch.md
```diff
@@ -0,0 +1,392 @@
...
+
+The syntax for single term matching is simply the text of the term itself. If one character of the term is an asterisk ```*``` it can match any number of characters forming a wildcard search. If the field-index does not have the suffix tree, then the position of the wildcard is restricted to only the end of the term.
+
+#### Phrase Matching
+
+Phrase matching is specified by enclosing a sequence of terms in a pair of double quotation marks ```"```. Phrase matching is only possible when term-offsets are specified in the index.
+
+### Limits
+
+To avoid combinatorial explosion certain operations have configurable limits applied.
```
madolson: What are the limits?

madolson (commented):
Did a quick read to ask questions I didn't follow. @stockholmux for your reference.

ValkeyFullTextSearch.md
```diff
@@ -0,0 +1,392 @@
...
+# Title
+Text Searching
+
+## Abstract
+
+The existing search platform is extended with a new field type, ```TEXT``` is defined. The query language is extended to support locating keys with text fields based on term, wildcard, fuzzy and exact phrase matching.
+
+## Motivation
+
+Text searching is useful in many applications.
```
madolson, line 15: We should identify the persona here. This is one of the most important questions to get the use case right. Is it for caching, primary database, etc.

ValkeyFullTextSearch.md
```diff
@@ -0,0 +1,392 @@
...
+│                                                       │ Term Storage in    │  │
+│                                                       │ Prefix/Suffix Trees│  │
+│                                                       └────────────────────┘  │
+│                                                                               │
+└───────────────────────────────────────────────────────────────────────────────┘
+
+```
+## Specification
+
+Each ```TEXT``` field has the a set of configurables some control the process to convert a string into a vector of terms, others control the contents of the generated index.
```
madolson:
```suggestion
Each ```TEXT``` field has the a set of configurations that control the process to convert a string into a vector of terms, others control the contents of the generated index.
```

ValkeyFullTextSearch.md
```diff
@@ -0,0 +1,392 @@
...
+
+| Metadata          | Type              | Meaning                                                                    | Default Value |
```
madolson, line 362: There is some slight inconsistency between this table and the later values, just make sure they're consistent.

ValkeyFullTextSearch.md
andrross: +1

Coming from the OpenSearch project, the potential scope of this RFC is enormous. OpenSearch leverages Lucene for most of the search functionality described here. Except for stemming, I don't see any other libraries mentioned, so does that mean you plan to implement everything else? More details about the targeted persona and use case could help narrow the scope.

msfroh:
Reading through the RFC, I'm curious about the implementation of the term dictionary. 

Will it just be a key/value lookup from full terms to postings lists? (I guess that would fit in well to Valkey.) I guess each term is indexed in its forward and reversed representation in the prefix/suffix trees?

Lucene uses finite-state transducers to encode the term dictionary. Similarly, Rust has the `fst` crate which is used for full-text search in LanceDB and (with some modifications) in the Tantivy library. Just wondering if there's a plan to build a similar FST solution for Valkey. It makes the dictionary pretty compact, while more efficiently handling wildcard/fuzzy matches. I don't know if there are good off-the-shelf FST libraries in C (but I wouldn't be surprised).

ValkeyFullTextSearch.md
```diff
@@ -0,0 +1,392 @@
...
+| stemming    | A process of mapping similar words into a common base word. For example, the words _drives_, _drove_ and _driven_ would be replaced with the  _drive_.                                                                      |
+| term        | The output of stemming a word.                                                                                                                                                                                              |
+| prefix tree | A mapping from input string to an output object. Insert/delete operations are O(length(input)). Additional operations include the ability to iterate over the entries that share a common prefix in O(length(prefix)) time. |
+
+## Design considerations
+
+The text searching facility provides machinery that decomposes text fields into terms and field-indexes them.
+The query facility of the ```FT.SEARCH``` and ```FT.AGGREGATE``` commands is enhanced to select keys based on combinations of term, fuzzy and phrase matching together with the existing selection operators of TAG, Numeric range and Vector searches.
```
stockholmux: nit: why is tag all caps?

ValkeyFullTextSearch.md
```diff
@@ -0,0 +1,392 @@
...
+
+### Tokenization process
+
+A tokenization process is applied to strings of text to produce a vector of terms.
```
stockholmux: nit: slightly confusing use of 'vector' here given the vector _search_ capabilities.

ValkeyFullTextSearch.md
```diff
@@ -0,0 +1,392 @@
...
+Tokensization is applied in two contexts. 
+First as part of the ingestion process for text fields of keys.
+Second to process query string words and phrases.
+
+The tokenization process has four steps.
+
+1. The text is tokenized, removing punctuation and redundant whitespace, generating a vector of words.
+2. Latin upper-case characters are converted to lower-case.
```
stockholmux: This feels like a miss. What about other scripts with upper/lower case? Cyrillic? Greek?

ValkeyFullTextSearch.md
```diff
@@ -0,0 +1,392 @@
...
+                                                                           V
+                                                                      ┌───────────────┐
+                                                                      │ Stemming      │
+                                                                      │ ["run",       │
+                                                                      │  "search",    │
+                                                                      │  "cat"]       │
+                                                                      └───────────────┘
+```
+
+Step 1 of the tokenization process is controlled by a specified list of punctuation characters.
```
stockholmux, line 97: is the specified list customizable?

ValkeyFullTextSearch.md
```diff
@@ -0,0 +1,392 @@
...
+Step 1 of the tokenization process is controlled by a specified list of punctuation characters. 
+Sequences of one or more punctuations characters delimit word boundaries.
+Note that individual punctuation characters can be escaped using the normal backslash prefix syntax, causing the next character to be treated as non-punctuation.
+
+### Text Field-Index Structure
+
+The text field-index is a mapping from a term to a list of locations.
+Depending on the configuration, the list of locations may be just a list of keys or a list of key/term-offset pairs. The list of locations is also known as a postings or a postings list.
+
+The mapping structure is built from one or two prefix trees. When present, the second prefix tree is built from each inserted term sequenced in reverse order, effectively providing a suffix tree.
```
stockholmux, line 106: how does this work for LTR and RTL languages?

ValkeyFullTextSearch.md
```diff
@@ -0,0 +1,392 @@
...
+
+Here is a visual example. Note that in the drawing, the sections noted as "same postings as _xxxx_" mean that the pointer actually points to the same memory area as referenced. In other words, when a suffix tree is present, the postings list is referenced by both (think shared_ptr).
+
+```
+┌──────────────────────────────────────────────────────────────────────────────┐
```
stockholmux, line 111: It feels like there should be a better way of achieving this without duplicating the data structure for prefix and suffix.

ValkeyFullTextSearch.md
```diff
@@ -0,0 +1,392 @@
...
+```
+
+
+### Search query operators
+
+Unlike the Vector, Tag and Numeric search operators the specification of a field is optional. If the field specification is omitted then a match is declared if any text field matches the search operator.
+
+#### Term matching
+
+There are three types of term matching: exact, wildcard and fuzzy. Exact term matching is self-descriptive, i.e., only keys containing exactly the specified text are matched.
```
stockholmux: Wait. Does exact match _everything_ before pre-processing (stemming, etc.) or just terms? I wouldn't gloss over this in the RFC - too easy to make assumptions.

ValkeyFullTextSearch.md
```diff
@@ -0,0 +1,392 @@
...
+
+Wildcard matching provides a subset of reg-ex style matching of terms.
```
stockholmux: Would glob style be more accurate or are you doing more (RegEx would be... complex)

ValkeyFullTextSearch.md
```diff
@@ -0,0 +1,392 @@
...
+Initially, only a single wildcard specifier ```*``` is allowed which matches any number of characters in a term.
+The wildcard can be at any position within the term, providing prefix, suffix and infix style matching.
+Note, in this proposal, the second prefix tree is required to perform infix and suffix matching.
+
+#### Fuzzy matching
+
+Fuzzy word matching is specified by enclosing a word in pairs of percent signs ```%```. This term matches words within one [Levenshtein edit distance](https://en.wikipedia.org/wiki/Levenshtein_distance) edit distance for each pair of percent signs from the specified word.
```
stockholmux: Consider making this pluggable. Levenshtein has weaknesses and there are many algorithms that suit different use cases better.

ValkeyFullTextSearch.md
```diff
@@ -0,0 +1,392 @@
...
+[WITHOFFSETS | NOOFFSETS]
+```
+
+Controls whether term-offsets are tracked in the field-index. Default is ```WITHOFFSETS```.
+
+```
+[NOSTOPWORDS | STOPWORDS <count> [word ...]]
+```
+
+Controls the application and list of stop words. Note that ```STOPWORDS 0``` is equivalent to ```NOSTOPWORDS```. The default stop words are:
```
stockholmux: This part of the API gives me a headache - the english only orientation of this API is super annoying to those who need indexes in other langauges.

A big strength of having a search engine in a database like Valkey is that you can create adhoc indexes rapidly. However, if you're doing it non-english languages you a have to specify the stop words on every index creation shifting complexity to the user. I would find a list of stop words for the stemming supported languages and adding them to the codebase.

Something like `STOPWORDSLANG "hindi"` to get all the default hindi stop words.

stockholmux: Additionally, there doesn't seem to be a way to augment the default stop words, so, say, I want to add the stop word 'poop' to the default stop words, I have to re-add _all_ the stop words. So I have to do 


```
... STOPWORDS 34 a is the an and are as at be but by for if in into it no not of on or such that their then there these they this to was will with poop ..
```

There should be some capacity to do `STOPWORDSADD 1 poop`

ValkeyFullTextSearch.md
```diff
@@ -0,0 +1,392 @@
...
+
+```
+[NOSTEM | LANGUAGE <language>]
+```
+
+Controls the stemming algorithm. Supported languages are: 
+
+**none, arabic, armenian, basque, catalan, danish, dutch, english, estonian, finnish, french, german, greek, hindi, hungarian, indonesian, irish, italian, lithuanian, nepali, norwegian, porter, portuguese, romanian, russian, scripts, serbian, spanish, swedish, tamil, turkish, yiddish**
+
+The default language is **english**. Note: ```LANGUAGE none``` is equivalent to ```NOSTEM```.
```
stockholmux, line 406: How do you handle indexes with text in multiple languages? It's a super common problem.

e.g.
document 1 is in English
document 2 is in Italian

If you try to index anything like that you're stems will be all over the place.

ValkeyFullTextSearch.md
```diff
@@ -0,0 +1,392 @@
...
+
+| Metadata          | Type              | Meaning                                                                    | Default Value |
+| ----------------- | ----------------- | -------------------------------------------------------------------------- | ------------- |
+| Punctuation       | vector<character> | Characters that separate words during the tokenization process.            |
+| Stop Words        | vector<String>    | List of stop words to be removed during decomposition                      |
+| Stemming Language | Enumeration       | Snowball stemming library language control                                 |
+| Suffix Tree       | Boolean           | Controls the presence/absence of a suffix tree.                            |
+| Word Offsets      | Boolean           | Indicates if the postings for a word contain word offsets or just a count. |
+
+### Extension of the ```FT.CREATE``` command
```
stockholmux, line 370: So schema changes require a `FT.DROP` and complete re-index?

ValkeyFullTextSearch.md
```diff
@@ -0,0 +1,392 @@
...
+RFC: 24
+Status: (Change to Proposed when it's ready for review)
+---
+
+# Title
+Text Searching
+
+## Abstract
+
+The existing search platform is extended with a new field type, ```TEXT``` is defined. The query language is extended to support locating keys with text fields based on term, wildcard, fuzzy and exact phrase matching.
```
stockholmux, line 11: So, using this type of query engine for full text search is next to impossible without `FT.EXPLAIN` what are the plans for this command?

ValkeyFullTextSearch.md
```diff
@@ -0,0 +1,392 @@
...
+### Extension of the ```FT.CREATE``` command
+
+Clauses are provided to control the configuration. 
+If supplied before the ```SCHEMA``` keyword then the default value of the clause is changed for any text fields declared in this command.
+If supplied as part of an individual field declaration, i.e., after the ```SCHEMA``` keyword, then it sets the configurable for just that field.
+
+```
+[PUNCTUATION <string>]
+```
+The characters of this string are used to split the input string into words. Note, the splitting process allows escaping of input characters using the usual backslash notation. This string cannot be empty. Default value is: _TBD_
```
stockholmux: This is language dependent, so like the problems global language defaults in stopwords (see below), the default punctuation should probably be linked with stemming language

ValkeyFullTextSearch.md
```diff
@@ -0,0 +1,392 @@
...
+| stemming    | A process of mapping similar words into a common base word. For example, the words _drives_, _drove_ and _driven_ would be replaced with the  _drive_.                                                                      |
+| term        | The output of stemming a word.                                                                                                                                                                                              |
+| prefix tree | A mapping from input string to an output object. Insert/delete operations are O(length(input)). Additional operations include the ability to iterate over the entries that share a common prefix in O(length(prefix)) time. |
+
+## Design considerations
+
+The text searching facility provides machinery that decomposes text fields into terms and field-indexes them.
+The query facility of the ```FT.SEARCH``` and ```FT.AGGREGATE``` commands is enhanced to select keys based on combinations of term, fuzzy and phrase matching together with the existing selection operators of TAG, Numeric range and Vector searches.
+
+### Tokenization process
```
stockholmux, line 58: How does this work with CJK and similar languages?

ValkeyFullTextSearch.md
```diff
@@ -0,0 +1,392 @@
...
+
+This clause controls the minimum number of characters in a word before it is subjected to stemming. Default value is 4.
+
+```
+[NOSTEM | LANGUAGE <language>]
+```
+
+Controls the stemming algorithm. Supported languages are: 
+
+**none, arabic, armenian, basque, catalan, danish, dutch, english, estonian, finnish, french, german, greek, hindi, hungarian, indonesian, irish, italian, lithuanian, nepali, norwegian, porter, portuguese, romanian, russian, scripts, serbian, spanish, swedish, tamil, turkish, yiddish**
```
stockholmux: miss: hebrew. Consider something pluggable in cases where Snowball doesn't cover.

stockholmux (changes requested):
Looks like there are some localization-related issues.

ValkeyFullTextSearch.md
allenss-amazon: yes

ValkeyFullTextSearch.md
allenss-amazon: Default values are TBD.

ValkeyFullTextSearch.md
allenss-amazon: Yes, the plan is to implement the necessary search structures in Valkey code. 

This implementation distinguishes itself from OpenSearch primarily in the time/space continuum. In that OpenSearch is more balanced toward saving space while this implementation focusses on low latency. We expect dramatically shorter query and ingestion times than what OpenSearch supports. In general, this is targetting a more dynamic corpus, i.e., very short latency from an insertion until it's visible in a query.

allenss-amazon:
FST data structure was examined and rejected due to the high run-time costs. Applications like OpenSearch that are FST based use a log-structure-merge update technique that causes both significant write-amplification as well as longer query operations. This choice of more time efficient (but less space efficient) data structures will be how this implementation distinguishes from OpenSearch (and equivalents)

ValkeyFullTextSearch.md
allenss-amazon: Use cases will differentiate based on the fundamentals:

1. Lower $/query. Due to time-optimized data structures and reduced parsing/command processing overhead.
2. Incremental update machinery. Data structures support incremental updates without GC. 
3. Current queries, i.e., query operations include the most recent mutations, i.e., mutation to query visibility in less than a milisecond.

allenss-amazon:
@msfroh (Sorry, the comment is located here, github wouldn't let me comment directly on your comment).

Yes, the intent is to implement word dictionaries using forward and reverse direction radix trees. Each word entry in the dictionary references a full postings list.

ValkeyFullTextSearch.md
allenss-amazon: Support for FT.EXPLAIN makes sense.

ValkeyFullTextSearch.md
allenss-amazon: Yes. Right now, there are no plans to mutate existing index definitions.

ValkeyFullTextSearch.md
allenss-amazon: Not sure that LTR vs RTL makes a difference here. Options like $inorder and $slop are trivially sensitive to ordering.

ValkeyFullTextSearch.md
allenss-amazon: Yes.

ValkeyFullTextSearch.md
allenss-amazon: Will reword, the intention is to be case-insensitive.

allenss-amazon: Would it make sense to have an option control this?

ValkeyFullTextSearch.md
madolson: Include that and at least include how these limits will be derived.

ValkeyFullTextSearch.md
KarthikSubbarao: We are planning to support english first, and the tokenization process can be extended in the future to support other languages

ValkeyFullTextSearch.md
KarthikSubbarao: I can include this detail. We are only planning on supporting english in the initial version, but we can make a note that hebrew needs a different handling.

ValkeyFullTextSearch.md
KarthikSubbarao: Yes, by current definition, this is more of a simple glob style. I can update the wording

ValkeyFullTextSearch.md
KarthikSubbarao: In the implementation, we are planning on only using the Suffix Tree when needed and is configurable through `[WITHSUFFIXTRIE | NOSUFFIXTRIE]`.

Other ways include using FSA/FST (finite-state transducers) to support both directions. As @allenss-amazon mentioned here, https://github.com/valkey-io/valkey-rfc/pull/24#issuecomment-3091220247, this comes with additional run time cost. So, we did not go with this option


An index structure which is similar to the RFC's proposed dual tree (prefix/suffix) approach would be a Edge-Sharing Dual Trie. This can optimize additionally (beyond just compression of trees) by reducing nodes since we use a single tree structure. To use this, we need to have the same APIs and same characteristics. I will validate if this is possible

ValkeyFullTextSearch.md
KarthikSubbarao: I will update the ordering to match the table's order with what follows

ValkeyFullTextSearch.md
KarthikSubbarao: This is a good point. Currently, only the indexes have a language specification. The documents (hash or JSON) do not natively provide clarity on the language used within it.

For Multilingual documents, we have two options:

1. the user could set LANGUAGE to none (disable stemming and stop word removal). We can add this to recommendations and best practices. 

2. The user can use a "Tag" for filtering documents which certain languages, as Allen mentioned.

ValkeyFullTextSearch.md
KarthikSubbarao: Some of the limits are already implemented in valkey-search and not specific to full text search. (
Query nesting depth, number of nodes in the query syntax tree, etc).


Some full text specific limits (other than what is below) are: MAXEXPANSIONS, MAXPREFIXEXPANSIONS.


We can also consider adding a limit for MAXSEARCHRESULTS

ValkeyFullTextSearch.md
allenss-amazon: I think the default stopwords should be language specific. Add commands to differentially add/remove from the default list is easy to do.

ValkeyFullTextSearch.md
allenss-amazon: The current plan is that the language is per-index. A future-feature of index definition is to add the FILTER option, which would allow dynamic selection of index membership based on prefixes AND an arbitrary expression. So if, for example, in your example, document 1 had a field marked "English" and document 2 had that same field marked as "Italian", then you could create an index with language English, containing JUST english documents. LIkewise a second index of Italian documents could contain just those documents.

Once aggregation gets merged, it provides the primitives to make the FILTER expression simple to implement.

ValkeyFullTextSearch.md
allenss-amazon: Agreed, I believe that ultimately the input tokenizer will have substantial per-language dependent functionality.

ValkeyFullTextSearch.md
KarthikSubbarao: > using this type of query engine for full text search is next to impossible without FT.EXPLAIN what are the plans for this command?

Basic text queries can be simple to understand. Yes, they can get complex, but so can hybrid vector queries (without text). 

FT.EXPLAIN provides value for all these cases, so it makes sense to support it. We can work towards adding support for it

ValkeyFullTextSearch.md
KarthikSubbarao: > We should identify the persona here.


Thanks everyone. I will update this section with more details

ValkeyFullTextSearch.md
KarthikSubbarao: I am taking a look

KarthikSubbarao: I can update this to clarify that exact matching operates on the normalized terms after pre-processing (tokenization, lowercasing, stemming if enabled). It does not match the raw input string exactly.

rikkarth:
Hey I'm currently using valkey (AWS) as the backend for my "redis-like" cache in an A.I. related project where valkey acts as the memory cache of this agent.

Because we are using `langgraph`, we thought we could use https://github.com/redis-developer/langgraph-redis, but using this library with a Valkey backend proves to be incompatible.

So we had to implement our own custom BaseCheckpointSaver (see https://github.com/redis-developer/langgraph-redis), since the redis one isn't compatible with Valkey's API.

First error we had was on `redisconnectionRedisConnectionFactory.validate_sync_redis` has been workarounded.

Then other errors were raised after connection `redis.exceptions.ResponseError: unknown command 'ft._list''`

recommended to post here by @madolson 

Thank you so much, would be amazing if we could make this work for Valkey!

# Description

```python
from langgraph.checkpoint.redis import RedisSaver

with RedisSaver.from_conn_string("redis://localhost:6379") as checkpointer:
    return checkpointer
```

## Produces

```bash
Exception Type: <class 'redisvl.exceptions.RedisModuleVersionError'>
Unexpected error occurred: Required Redis db module search >= 20600 OR searchlight >= 20600 not installed. See Redis Stack docs at https://redis.io/docs/latest/operate/oss_and_stack/install/install-stack/.
Traceback (most recent call last):
... 
  File "./test.py", line XX, in _redis_cache
    with RedisSaver.from_conn_string("redis://localhost:6379") as checkpointer:
         ~~~~~~~~~~~~~~~~~~~~~~~~~~~^^^^^^^^^^^^^^^^^^^^^^^^^^
  File "./.pyenv/versions/3.13.1/lib/python3.13/contextlib.py", line 141, in __enter__
    return next(self.gen)
  File "./lib/python3.13/site-packages/langgraph/checkpoint/redis/__init__.py", line 497, in from_conn_string
    saver = cls(
        redis_url=redis_url,
    ...<2 lines>...
        ttl=ttl,
    )
  File "./lib/python3.13/site-packages/langgraph/checkpoint/redis/__init__.py", line 55, in __init__
    super().__init__(
    ~~~~~~~~~~~~~~~~^
        redis_url=redis_url,
        ^^^^^^^^^^^^^^^^^^^^
    ...<2 lines>...
        ttl=ttl,
        ^^^^^^^^
    )
    ^
  File "./lib/python3.13/site-packages/langgraph/checkpoint/redis/base.py", line 125, in __init__
    self.configure_client(
    ~~~~~~~~~~~~~~~~~~~~~^
        redis_url=redis_url,
        ^^^^^^^^^^^^^^^^^^^^
        redis_client=redis_client,
        ^^^^^^^^^^^^^^^^^^^^^^^^^^
        connection_args=connection_args or {},
        ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
    )
    ^
  File "./lib/python3.13/site-packages/langgraph/checkpoint/redis/__init__.py", line 70, in configure_client
    self._redis = redis_client or RedisConnectionFactory.get_redis_connection(
                                  ~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~^
        redis_url, **connection_args
        ^^^^^^^^^^^^^^^^^^^^^^^^^^^^
    )
    ^
  File "./lib/python3.13/site-packages/redisvl/redis/connection.py", line 261, in get_redis_connection
    RedisConnectionFactory.validate_sync_redis(
    ~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~^
        client, required_modules=required_modules
        ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
    )
    ^
  File "./lib/python3.13/site-packages/redisvl/redis/connection.py", line 411, in validate_sync_redis
    validate_modules(installed_modules, required_modules)
    ~~~~~~~~~~~~~~~~^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
  File "./lib/python3.13/site-packages/redisvl/redis/connection.py", line 189, in validate_modules
    raise RedisModuleVersionError(error_message)
redisvl.exceptions.RedisModuleVersionError: Required Redis db module search >= 20600 OR searchlight >= 20600 not installed. See Redis Stack docs at https://redis.io/docs/latest/operate/oss_and_stack/install/install-stack/.
```

Adding this bellow fixes this issue
```python
from redisvl.redis.connection import RedisConnectionFactory

RedisConnectionFactory.validate_sync_redis = (
    lambda redis_client, lib_name=None, required_modules=None: None
)
with RedisSaver.from_conn_string(
    "redis://localhost:6379"
) as checkpointer:
    return checkpointer
```

Then

```bash
Exception Type: <class 'redisvl.exceptions.RedisSearchError'>
Error while searching: unknown command 'FT.SEARCH', with args beginning with: 'checkpoints' '(@thread_id:{15cb9aad\-5021\-4dbb\-a5cc\-2bb7a350c34e} @checkpoint_ns:{__empty__})' 'RETURN' '6' 'thread_id' 'chec' 
Traceback (most recent call last):
  File "./lib/python3.13/site-packages/redisvl/index/index.py", line 934, in search
    return self._redis_client.ft(self.schema.index.name).search(
           ~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~^
        *args, **kwargs
        ^^^^^^^^^^^^^^^
    )  # type: ignore
    ^
  File "./lib/python3.13/site-packages/redis/commands/search/commands.py", line 519, in search
    res = self.execute_command(SEARCH_CMD, *args, **options)
  File "./lib/python3.13/site-packages/redis/client.py", line 622, in execute_command
    return self._execute_command(*args, **options)
           ~~~~~~~~~~~~~~~~~~~~~^^^^^^^^^^^^^^^^^^
  File "./lib/python3.13/site-packages/redis/client.py", line 633, in _execute_command
    return conn.retry.call_with_retry(
           ~~~~~~~~~~~~~~~~~~~~~~~~~~^
        lambda: self._send_command_parse_response(
        ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
    ...<2 lines>...
        lambda _: self._close_connection(conn),
        ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
    )
    ^
  File "./lib/python3.13/site-packages/redis/retry.py", line 87, in call_with_retry
    return do()
  File "./lib/python3.13/site-packages/redis/client.py", line 634, in <lambda>
    lambda: self._send_command_parse_response(
            ~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~^
        conn, command_name, *args, **options
        ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
    ),
    ^
  File "./lib/python3.13/site-packages/redis/client.py", line 605, in _send_command_parse_response
    return self.parse_response(conn, command_name, **options)
           ~~~~~~~~~~~~~~~~~~~^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
  File "./lib/python3.13/site-packages/redis/client.py", line 649, in parse_response
    response = connection.read_response(disable_decoding=True)
  File "./lib/python3.13/site-packages/redis/connection.py", line 666, in read_response
    raise response
redis.exceptions.ResponseError: unknown command 'FT.SEARCH', with args beginning with: 'checkpoints' '(@thread_id:{15cb9aad\-5021\-4dbb\-a5cc\-2bb7a350c34e} @checkpoint_ns:{__empty__})' 'RETURN' '6' 'thread_id' 'chec' 

The above exception was the direct cause of the following exception:

Traceback (most recent call last):
  ...
  File "./lib/python3.13/site-packages/langgraph/checkpoint/redis/__init__.py", line 361, in get_tuple
    results = self.checkpoints_index.search(checkpoints_query)
  File "./lib/python3.13/site-packages/redisvl/index/index.py", line 942, in search
    raise RedisSearchError(f"Error while searching: {str(e)}") from e
redisvl.exceptions.RedisSearchError: Error while searching: unknown command 'FT.SEARCH', with args beginning with: 'checkpoints' '(@thread_id:{15cb9aad\-5021\-4dbb\-a5cc\-2bb7a350c34e} @checkpoint_ns:{__empty__})' 'RETURN' '6' 'thread_id' 'chec' 
```

allenss-amazon:
This would be more appropriate in the valeky-search repo issues. which it looks like it was referenced in.

ValkeyFullTextSearch.md
```diff
@@ -0,0 +1,402 @@
...
+                                                                      ┌───────────────┐
+                                                                      │ Stemming      │
+                                                                      │ ["run",       │
+                                                                      │  "search",    │
+                                                                      │  "cat"]       │
+                                                                      └───────────────┘
+```
+
+Step 1 of the tokenization process is controlled by a specified list of punctuation characters. 
+Sequences of one or more punctuations characters delimit word boundaries.
```
sachinvmurthy, line 98: How will hyphenated words behave ("batman-dark-knight-rises")?
Will "batman", "dark", "knight", "rises" be separate? Or should certain punctuation be considered “in-word” optionally?

ValkeyFullTextSearch.md
allenss-amazon: One use-case that I have seen repeated is to search session histories. In this use-case, per-user history is loaded when a session is initiated (either a web-session or sometimes a "log on" of an application).

Another use-case is in RAG operations with filters that have additional text phrasing -- phrasing that can't be predicted and turned into a TAG operation.

ValkeyFullTextSearch.md
KarthikSubbarao: Yes, it will tokenized into "batman", "dark", "knight", "rises" be separate in ingestion IF it is a punctuation character (by default it is) specified in the index creation. 

On the query side, if - is a punctuation character, in an exact phrase, it is simply stripped out. If it is not an exact phrase, it will be parsed as: Term(batman), Negate(Term(dark)), Negate(Term(knight)), Negate(Term(rises)). This is because - is a query syntax character indicating negation. The user can escape this by using the backslash if needed, if their punctuation does not include -, and they want the query to match the ingestion for searches to match

PR #25, opened 2025-08-14
[DROPPED] add RFC for Key-Value option of External Data Storage

kronwerk:
this is a RFC initiative for https://github.com/valkey-io/valkey/issues/83
2 RFCs are proposed, both relating to External Data Storage module allowing to store data on persistent storage (mostly disks, I guess) in addition to current memory data storage:
1) Key-value pair storage (this PR);
2) Values only storage (number 26).

Different RFCs are not mentioned to be competitive in some way - in my opinion those options could perfectly be both implemented and selected to be used by a single Valkey config option. By they are just restrictive to be used on a single database simultaneously and have rather specific differences in implementation, that's why those RFCs are separated.

As a best effort of this initiative I see both RFCs are Approved after some discussion - as currently I see no problems to implement both in a rather non-interfering way combined with reusing some common code.

### DIFF
--- a/ExtDataKeyValue.md
+++ b/ExtDataKeyValue.md
@@ -1,0 +1,167 @@
+---
+RFC: 25
+Status: Proposed
+---
+
+# Title
+External data storage (keys + values)
+
+## Abstract
+Currently Valkey allows to store data in memory only. The option to store data externally (on local or network disk, e.g. SSD) is proposed here.
+Both keys and values are about to be stored externally - option to store values only externally is proposed in another RFC (#26).
+
+## Motivation
+1) Cost saving - disks are cheaper than memory.
+2) Larger storage - memory ends up fast in some use cases.
+3) NOT a data durability enhancement - this is out of scope of this initiative.
+
+## Design considerations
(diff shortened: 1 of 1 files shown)


kronwerk:
@madolson let's discuss this here: https://github.com/valkey-io/valkey-rfc/pull/26#discussion_r2289471239

> we probably want a more concrete implementation

you mean actually working and fully functional external storage module for this KV-option to take a look at its actual implementation to decide whether it's ok to Approve this current RFC? or I've misunderstood you, pls clarify a bit

kronwerk:
@madolson in the context of your comment mentioned above there's another thing - about @ferbs suggestion: https://github.com/valkey-io/valkey-rfc/pull/26#issuecomment-3193915989

does it mean that this is not an optimization option for Keys-Values/RFC from your perspective? storing only dict with hashes in Valkey instead of actual keys means that we can't perform scan/keys without accessing external storage from Valkey, and we might possibly evict hashes after some additions to eviction mechanism but we need some unobvious way to cleanup respective K/V pair from external storage (or leave it till TTL gonna clean it up which looks resource wasteful)

kronwerk:
@PingXie correct me if I'm wrong - as I believe you were interested in an option of Key and Values in external storage (or on disk, to simplify) instead of storing Values only: https://github.com/valkey-io/valkey/issues/83#issuecomment-2527065529 and https://github.com/valkey-io/valkey/issues/83#issuecomment-2537981963

could you please take a look at madolson@ comment above and add some considerations about that if you have any

PR #26, opened 2025-08-14
add RFC for External Data Storage

kronwerk:
this is a RFC initiative for https://github.com/valkey-io/valkey/issues/83
2 options inside a single RFC are proposed, both relating to External Data Storage module allowing to store data on persistent storage (mostly disks, I guess) in addition to current memory data storage:
1) Key-value pair option;
2) Values-only option.

Different options are not mentioned to be competitive in some way - in my opinion those options could perfectly be both implemented and selected to be used by a single Valkey config option. They are just restrictive to be used on a single database simultaneously and have rather specific differences in implementation, that's why things different for those options are separated in this RFC.


ferbs:
> Keys are just stored as usual

A potential middle ground compromise between this RFC and #25 would be keeping in memory a non-crypto hash of external-storage keys without holding the full key itself in memory. It could instead keep a ref count of externally stored keys by hashed value. This makes hash collisions of low consequence, letting you choose a fast algorithm with minimum output size [(eg, rapidhash, a5hash, other?)](https://github.com/rurban/smhasher)

To summarize the differences, moving out both keys and values (26) saves the most memory at the cost of performance because it checks external storage too often. Moving out just values (25) avoids these IO delays but potentially consumes too much memory. The non-crypto hash approach mostly avoids the IO delays, but still consumes some memory and adds some cpu load.

ferbs:
> The whole thing on this stage is about strings only - it's not about the impossibility or lack of demand for other Valkey data structures (sets, hashes, etc.), it's about the scope.

I'm wondering if different initial feature choices might deliver on more of the use cases described in https://github.com/valkey-io/valkey/issues/83. Starting with strings-only would primarily support only that large-blob case. 

Would deferring most or all of the user config options and command flags simplify the scope enough to implement more Valkey data structures in the first pass? 

The granular user options in this RFC are desirable but some of the high-impact ideas mentioned in that data-tiering discussion would still be useful without them. A minimalist hot-to-cold approach might look like an existing eviction policy but instead of deletion, data is moved to external storage. That alone would be a big feature and would support important use cases mentioned in the discussion: streams, reducing out-of-memory buckling, and large-ish blobs.

kronwerk:
> I'm wondering if different initial feature choices might deliver on more of the use cases described in [valkey-io/valkey#83](https://github.com/valkey-io/valkey/issues/83). Starting with strings-only would primarily support only that large-blob case.

those are Valkey data types:
- [X]String
- Hash
- List
- Set
- Sorted set
- [X]Stream
- Bitmap
- Bitfield
- Geospatial

@ferbs I've checked Strings and Streams - are there another types from this list that you see provided from the very start?

kronwerk:
> A potential middle ground compromise between this RFC and #25 would be keeping in memory a non-crypto hash of external-storage keys

@ferbs I've thought about that a bit - it is totally understandable why this could be useful, but (correct me if I'm wrong) it seems to be reasonable only with large keys. because otherwise we just get option #25 with additional complexity and collisions.

is "large keys" a valid and common use case enough to bother with that? I just don't know that, are there any considerations about how to determine that?

ferbs:
@kronwerk I see [max key length](https://valkey.io/topics/keyspace/) in Valkey is is 512 MB. Hahah, no idea who's doing that but I'd bet they do get large. Building a key by joining unbound variables is a fairly common pattern. Supporting large keys seems like a good idea, or at least not something to punt without compelling reason. Ok, so less complexity around collisions... another idea comes to mind:

Valkey keeps a set of 16 or 32 bit checksums for keys moved to external storage, with the burden of cleanup put on the module. Eg: "GET some-key" is a miss in memory; external storage module also has a miss; module checks its privately managed data structure of checksums, sees no collisions; module replies that the checksum should be removed from memory. Side note: the downside of a non-complying module is relatively low, extra checksums in memory are fairly small and the wasteful/avoidable io is limited to attempts to get deleted keys.

This should result in performance approaching rfc-25 and still save nearly as much memory as the no-key case. 

Another consideration comes to mind though. If you think about hardware crashes at various points of set or del operations, would one rfc offer any advantages over the other?

ferbs:
> are there another types from this list that you see provided from the very start?

My personal use case needs Hash. I can't really rationalize it over another on the list though.

ExtDataValuesOnly.md
```diff
@@ -0,0 +1,163 @@
...
+
+### FLUSHALL/FLUSHDB/SWAPDB flow
+1) Requires storing different databases externally separately enough to quickly drop|swap its data.
+
+### DBSIZE flow
+1) Requires value dynamic counter for each database to quickly return it without recounting on each request.
+
+### Backup flow
+1) External storage data should not be incorporated into RDB - this allows running Valkey core and external storage with its underlying data on different disks, either local or network ones.
+2) RDB and external storage backup should be synchronized to match keys in memory with external values - thus running (BG)SAVE in core should initiate DUMP command for local external storage, which should be performed only after currently running values changing operations are done and not before starting new ones (some lock seems to be unavoidable here).
```
madolson: Why have two separate files for backups? Seems like it will make backup more complicated for other systems, although it's probably simpler for the external storage system.

ExtDataValuesOnly.md
```diff
@@ -0,0 +1,163 @@
+---
+RFC: 26
+Status: Proposed
+---
+
+# Title
+External data storage (values only)
+
+## Abstract
```
madolson: Overall I think I like this approach more than storing all keys/values on disk, I think this will be much simpler to implement overall. (Stuff like scan, eviction, etc will be hard to reason about if keys are on disk). 

Abstracting everything to the storage layer feels a bit hand wavy atm, we probably want a more concrete implementation to validate this design against.

ExtDataValuesOnly.md
```diff
@@ -0,0 +1,163 @@
...
+- Valkey expiration mechanism doesn't touch external data anyhow.
+- The maximum living time since last read|write for value external storage could be set in Valkey config (VC2), it's existence is not guaranteed after that (0=don't gc ever - only by direct DEL command, -1=gc any time).
+- Value is not supposed to be returned back to memory from external storage on any occasion - this could be introduced separately later with some config options if that's really a latency problem for somebody in some usecases (doing that naively and by default could lead to ping-pong between memory and external storage slowing down database in whole for most or even all keys).
+- Valkey key is not deleted right after value freeing initiated by external storage - it gets removed on accessing it (see GET flow).
+
+### Commands
+1. EXT_DATA DEBUG (see Debug mechanism lower)
+2. EXT_DATA LOADED: list previuosly loaded modules that implement external storage functionality (to get module names for INIT command)
+3. EXT_DATA INIT <db> <module>: initialize previously loaded module with name <module> for a specified database with name <db>
```
madolson: Pushing all this initiation on the startup feels inconsistent with how we do other configurations, which is normally done at startup with config files. We can have these commands, but the main users should probably configure this at startup.

ExtDataValuesOnly.md
```diff
@@ -0,0 +1,163 @@
...
+
+## Specification
+VC<N> = Valkey config option with its number for reference between RFC parts
+VF<N> = flag for existing Valkey command with its number for reference between RFC parts
+* Passing value storage<->client should be made in a way to prevent OOM during loading large data to limited memory.
+
+### GET flow
+1) Client requests a key.
+2) Valkey dict contains the key with attribute "memory-stored" OR missing dict when requested without flag (VF1) - end of story.
+3*) Key attribute is "external-stored" - value is requested from external storage.
```
madolson:
```suggestion
3*) Key attribute is "external-stored" - value is requested from external storage.
```
How is it requested and delivered from external storage? Will we hand off the client to the module? We can't block the main thread.

ExtDataValuesOnly.md
kronwerk: 1. single file means storing on a single data drive - this could be inconvenient in some infrastructure options, e.g. dividing Valkey and external storage completely and interact each other via network.
2. single file means consequtive access or some complicated ways of parallel access for Valkey and external storage to the same file.
3. option of limiting access to RDB to Valkey only means that we need to transfer data from external storage (not likely a good option) or allow Valkey accessing external storage data files directly (looks like a architecture choice with some side effects).
4. changes in Valkey could break up the way external storage interact with RDB - and vice versa, external storage changes could mess up the RDB which is potentially more problematic. with several external storages in wildlife which is expected (and even desired IMHO) this could imply additional synchronization and testing requirements between Valkey and modules.

ExtDataValuesOnly.md
kronwerk: you mean actually working and fully functional external storage module for this KV-option to take a look at its actual implementation to decide whether it's ok to Approve this current RFC? or I've misunderstood you, pls clarify a bit

ExtDataValuesOnly.md
kronwerk: ok - could you possibly send a Valkey code example of how this is done in a way you mentioned for another functionality/module?

ExtDataValuesOnly.md
kronwerk: client is not supposed to be interacted with module directly - async bio job for returning value to client and discarding key on miss is expected (or another solution requiring doing that asynchronously in a separate thread)

kronwerk:
> A potential middle ground compromise between Keys-Values and Only-Values options would be keeping in memory a non-crypto hash of external-storage keys without holding the full key itself in memory.

@madolson in the context of your comment: https://github.com/valkey-io/valkey-rfc/pull/26#discussion_r2289471239

does it mean that ferbs@'s proposed thing is not a possible optimization for KV-option from your perspective? storing only dict with hashes in Valkey instead of actual keys means that we can't perform scan/keys without accessing external storage from Valkey, and we might possibly evict hashes after some additions to eviction mechanism but we need some unobvious way to cleanup respective K/V pair from external storage (or leave it till TTL gonna clean it up which looks resource wasteful)

kronwerk:
@PingXie correct me if I'm wrong - as I believe you were interested in an option of Key and Values in external storage (or on disk, to simplify) instead of storing Values only: https://github.com/valkey-io/valkey/issues/83#issuecomment-2527065529 and https://github.com/valkey-io/valkey/issues/83#issuecomment-2537981963

could you please take a look at madolson@ comments above and add some considerations about that if you have any

ferbs:
Maybe a useful exercise is thinking of the feature as eviction-tiering more than data-tiering, where Valkey doesn't strictly need to maintain the facade of a single unified data store. That would permit implementing the feature with command-level granularity, where a command is run first against the in-memory store, with the external store given a chance to append only after when relevant. Also, the entire key and value for any data structure is moved as a unit, evicted from memory to the external store. (eg, hset never splits items across stores.)

It seems to meet the primary goals, of improving the Valkey story for large data and/or rarely accessed data. Would it simplify implementation? I'm not familiar with the codebase, just speculating and trying to contribute, but it seems like it would work fine for scan (and eval too) at least from a superficial surface glance.

kronwerk:
@madolson @PingXie @zuiderkwast hi, I've updated the RFC +\- in accordance with my ongoing implementation of KV-approach mentioned in it.
it covers only core and module API changes, as I've mentioned earlier someday, not the storage module itself - a test storage module is implemented in some knowingly inefficient way for testing purposes only.
seems that it's about 80-85% ready now, so any comments/concerns/advices to RFC are still welcome - or you can just wait for this implementation attempt publish in next month (I hope), if it's more convenient to you. thx

PR #28, opened 2025-09-11
Create valkey-k8s-operator

andrey-glazkov:
This RFC proposes an open-source Kubernetes Operator for Valkey. It supports Cluster and Sentinel-based HA deployments and standalone nodes, with optional TLS, persistence, and Prometheus metrics integration.

Includes:
CRD schema and operator behavior
ConfigMap and Secret handling
Sentinel-managed shard support
AZ-aware scheduling design
Optional Failover CRD
Learnings from Helm-based deployments and internal prototypes


VALKEY-K8S-OPERATOR.md
```diff
@@ -0,0 +1,259 @@
+RFC: 21
+Status: Proposed
+
+Valkey Kubernetes Operator RFC
+
+Abstract
+
+This RFC proposes the development of an open-source Kubernetes Operator for Valkey. 
+The Operator will automate the provisioning, configuration, and lifecycle management of Valkey clusters within Kubernetes environments. 
+It will support standalone and Sentinel-based deployments, integrate with cert-manager for TLS, expose Prometheus metrics, and support persistent storage configurations, offering production-grade capabilities for cloud-native environments.
```
hpatro, line 10: Any reason leaving aside cluster-enabled mode ?

valkey-k8s-operator
```diff
@@ -0,0 +1,259 @@
...
+- Enabling self-service for developers via CRDs
+- Automating TLS setup, monitoring, and failover
+- Supporting advanced production use cases like Sentinel HA and persistent storage
+- Matching enterprise expectations for security, observability, and compliance
+
+Unlike Redis, which has various third-party operators with unclear support models, Valkey has the opportunity to lead with an officially supported Kubernetes-native experience.
+
+Design considerations
+
+The operator will be implemented using Kubebuilder and based on the Bitnami Helm chart for Valkey.
```
hpatro: I thought people want to move away from Bitnami Helm chart due to the recent pricing change.

VALKEY-K8S-OPERATOR.md
```diff
@@ -0,0 +1,259 @@
...
+Design considerations
+
+The operator will be implemented using Kubebuilder and based on the Bitnami Helm chart for Valkey. 
+It will expose a Custom Resource Definition (CRD) named `Valkey`, allowing users to declaratively deploy Valkey instances via YAML.
+
+Key design principles:
+- User-first: Apply 10 lines of YAML and get a working, HA Valkey deployment.
+- Composability: Work alongside tools like cert-manager, Prometheus Operator, and Everest.
+- Security: TLS and credential management must be default-safe.
+- Modularity: Start simple (standalone + Sentinel), then iterate (autoscaling, dashboards).
```
hpatro, line 43: Will be good to understand the availability difference between Standalone + Sentinel in K8S vs cluster-enabled setup.

valkey-k8s-operator
```diff
@@ -0,0 +1,259 @@
...
+  persistence:
+    enabled: true
+    type: both
+    storageClass: fast-disks
+    size: 5Gi
+  config:
+    configMapName: custom-valkey-conf
+  modules:
+    - name: redisbloom.so
+      path: /opt/valkey/modules/redisbloom.so
```
hpatro: Would prefer referring Valkey modules, even for example basis.

valkey-k8s-operator
```diff
@@ -0,0 +1,259 @@
...
+- Flexible deployment of masters, replicas, and Sentinel in separate StatefulSets
+- AZ-aware scheduling to avoid cross-AZ traffic and latency
+- Clear control over replica counts and placement
+- Optional integration of a custom controller for master awareness and controlled failover
+
+Design Overview
+
+Each Valkey Shard will be backed by multiple Kubernetes resources:
+
+- A StatefulSet for the master (1 pod, zone-pinned, highest priority)
```
hpatro:
```suggestion
- A StatefulSet for the primary (1 pod, zone-pinned, highest priority)
```

VALKEY-K8S-OPERATOR.md
bbarani: +1. It would be great to have cluster mode supported as well.

zuiderkwast: Cluster is the only deployment with HA and that scales. IMAO it's the only mode that's worth focusing on... For the simpler setups, a simple helm chart could be enough...

rlunar:
I would leave Sentinel as nice to have and Cluster Mode as a must have for version 0.1.0 besides Prometheus integration with [Coroot](https://github.com/coroot/coroot) will be ideal for Observability. Failover and node replacement has to be part of the functionality as well for reliability.

zuiderkwast:
> I would leave Sentinel as nice to have and Cluster Mode as a must have for version 0.1.0

I think we don't need to define exactly what's must-have for a particular version. We can line out the long term goals, implement things incrementally and just make sure we don't shut the door on any of these long term goals.

It's great to have something to start with. We should review this RFC and make it covers what people want.

VALKEY-K8S-OPERATOR.md
andrey-glazkov: good, updated motivation, design, specs based on this

valkey-k8s-operator
andrey-glazkov: right, my bad - that got here from old docs

valkey-k8s-operator
andrey-glazkov: oops :-) fixed!

VALKEY-K8S-OPERATOR
```diff
@@ -0,0 +1,267 @@
...
+Design considerations
+
+The operator will be implemented using Kubebuilder and expose a Custom Resource Definition (CRD) named `Valkey`, allowing users to declaratively deploy Valkey instances via YAML. 
+
+Key design principles:
+- User-first: Apply 10 lines of YAML and get a working, HA Valkey deployment.
+- Cluster-first focus: Prioritize cluster mode as the production-grade setup — HA, scalable, and resilient
+- Fallback / simpler modes: Support standalone and Sentinel HA deployments for small-scale or testing environments
+- Composability: Work alongside tools like cert-manager, Prometheus Operator, and Everest.
+- Security: TLS and credential management must be default-safe.
```
bjosv: TLS/mTLS ? but maybe that's a given

VALKEY-K8S-OPERATOR.md
```diff
@@ -0,0 +1,267 @@
...
+- Cluster-first focus: Prioritize cluster mode as the production-grade setup — HA, scalable, and resilient
+- Fallback / simpler modes: Support standalone and Sentinel HA deployments for small-scale or testing environments
+- Composability: Work alongside tools like cert-manager, Prometheus Operator, and Everest.
+- Security: TLS and credential management must be default-safe.
+- Modularity: Start simple (standalone + Sentinel), then iterate (autoscaling, dashboards).
+- Observability: First-class Prometheus support out-of-the-box.
+- Deployment approach: The operator will rely on either:
+  - A fully open-source Helm chart maintained internally
+  - Direct deployment manifests managed by the operator itself
+
```
bjosv, line 53: We should add something regarding
Reliability: mitigate failure scenarios by distribute data using anti-affinity, handle K8s worker failures, and rolling updates of workers without downtime.

VALKEY-K8S-OPERATOR.md
```diff
@@ -0,0 +1,267 @@
...
+  - Cluster mode deployments with multiple Sentinel-managed shards
+  - Persistent storage and backup/restore workflows
+- Providing simplified standalone or Sentinel-only deployments for smaller or test environments where full cluster mode is unnecessary
+- Matching enterprise expectations for security, observability, and compliance
+
+Unlike Redis, which has various third-party operators with unclear support models, Valkey has the opportunity to lead with an officially supported Kubernetes-native experience.
+
+Design considerations
+
+The operator will be implemented using Kubebuilder and expose a Custom Resource Definition (CRD) named `Valkey`, allowing users to declaratively deploy Valkey instances via YAML.
```
jdheyburn, line 33: I'm not really an operator expert, but I'm keen to see what the pros/cons are for kubebuilder vs operator-sdk.

VALKEY-K8S-OPERATOR
```diff
@@ -0,0 +1,267 @@
...
+Unlike Redis, which has various third-party operators with unclear support models, Valkey has the opportunity to lead with an officially supported Kubernetes-native experience.
+
+Design considerations
+
+The operator will be implemented using Kubebuilder and expose a Custom Resource Definition (CRD) named `Valkey`, allowing users to declaratively deploy Valkey instances via YAML. 
+
+Key design principles:
+- User-first: Apply 10 lines of YAML and get a working, HA Valkey deployment.
+- Cluster-first focus: Prioritize cluster mode as the production-grade setup — HA, scalable, and resilient
+- Fallback / simpler modes: Support standalone and Sentinel HA deployments for small-scale or testing environments
```
jdheyburn: Could we define what "standalone" means - self managed HA deployment w/o Sentinel?

VALKEY-K8S-OPERATOR
```diff
@@ -0,0 +1,267 @@
...
+- Composability: Work alongside tools like cert-manager, Prometheus Operator, and Everest.
+- Security: TLS and credential management must be default-safe.
+- Modularity: Start simple (standalone + Sentinel), then iterate (autoscaling, dashboards).
+- Observability: First-class Prometheus support out-of-the-box.
+- Deployment approach: The operator will rely on either:
+  - A fully open-source Helm chart maintained internally
+  - Direct deployment manifests managed by the operator itself
```
jdheyburn: Personally I would be in favour of direct deployment of manifests, but I guess it depends where the official Valkey helm chart is at the point of development.

VALKEY-K8S-OPERATOR.md
```diff
@@ -0,0 +1,267 @@
...
+    enabled: true
+    type: both
+    storageClass: fast-disks
+    size: 5Gi
+  config:
+    configMapName: custom-valkey-conf
+  modules:
+    - name: libvalkey_bloom.so
+      path: /opt/valkey/modules/libvalkey_bloom.so
+```
```
jdheyburn, line 134: As part of the design document, could we outline the primary CRDs we wish to create for future compatibility? i.e. _one `Valkey` CRD to rule them all_, or `ValkeyCluster`, `ValkeyStandalone`, etc.

Perhaps the all-in-one `Valkey` can create these CRDs depending on what `mode` is defined - then they are handled by their respective controllers.

VALKEY-K8S-OPERATOR.md
bjosv, line 1: We should rename this file to a markdown file for readablity (the repo readme states `file that's named in uppercase and ends in .md`), then probably add a table (`---`) and headings(`#`) as used in the [template](https://github.com/valkey-io/valkey-rfc/blame/main/TEMPLATE.md).

VALKEY-K8S-OPERATOR
```diff
@@ -0,0 +1,267 @@
...
+
+Operating Valkey at scale in Kubernetes environments remains complex and time-consuming. 
+Most enterprise users currently manage Valkey deployments manually or via generic Helm charts, lacking production-readiness, automation, or compliance guarantees.
+
+This Operator addresses these challenges by:
+- Providing a Kubernetes-native way to deploy and manage cluster mode Valkey, which is the only deployment type that provides HA and horizontal scaling.
+- Enabling self-service for developers via CRDs
+- Automating TLS setup, monitoring, and failover
+- Supporting advanced production use cases:
+  - Cluster mode deployments with multiple Sentinel-managed shards
```
zuiderkwast: Cluster and Sentinel are mutually exclusive. For a long time, I didn't realize this and I think it's confusing to a lot of people. They are two different ways to deploy Valkey and they can't be used together.

We should change this formulation. Probably split it into two separate bullets. 

Sentinel nodes are light-weight nodes without any data. They form a cluster and supervise other Valkey data nodes and trigger failover when they detect that a primary is down. A Sentinel cluster consists of at least 3 dedicated sentinel nodes that form a consensus cluster. A single Sentinel cluster can supervise multiple separate unsharded Valkey instances and their replicas. Unsharded here means that all keys belong to the same primary.

A Sentinel cluster can maybe be deployed independently from the data nodes. I believe we need a separate CRD just for deploying Sentinel clusters. Sentinel nodes are different to the data nodes, but in some setups, they are just supervising a single set of valkey nodes and then they can run in the same pods as the data nodes, or in side cars. The Valkey data nodes are not aware of the Sentinel nodes.

Cluster mode is a different mode for running the Valkey. In cluster mode, the data nodes them selves form the HA cluster and handle automatic failovers themselves. They also know the topology of the cluster and can redirects clients to the right nodes. A Valkey Cluster is always sharded with at least three shards. Clients/applications need be aware that they're connecting to a cluster. Cluster-aware clients hash the keys into *slots* using `slot(key) = CRC16(key) % 16384` and use this slot number to find out which primary owns each slot. This is what we mean by sharding and it's a model that allows to scale the cluster horizontally by adding more shards (a primary and some replicas) and moving some slots to the new primary. Let's not use the word "sharding" for Sentinel deployments, as it's easily confused with Cluster sharding.

```suggestion
  - Cluster mode deployments
  - Sentinel cluster deployments for supervising multiple primary-replica instances
```

VALKEY-K8S-OPERATOR
```diff
@@ -0,0 +1,267 @@
...
+The operator will be implemented using Kubebuilder and expose a Custom Resource Definition (CRD) named `Valkey`, allowing users to declaratively deploy Valkey instances via YAML. 
+
+Key design principles:
+- User-first: Apply 10 lines of YAML and get a working, HA Valkey deployment.
+- Cluster-first focus: Prioritize cluster mode as the production-grade setup — HA, scalable, and resilient
+- Fallback / simpler modes: Support standalone and Sentinel HA deployments for small-scale or testing environments
+- Composability: Work alongside tools like cert-manager, Prometheus Operator, and Everest.
+- Security: TLS and credential management must be default-safe.
+- Modularity: Start simple (standalone + Sentinel), then iterate (autoscaling, dashboards).
+- Observability: First-class Prometheus support out-of-the-box.
```
zuiderkwast: Does using redis-exporter in a side-car count as "first-class out-of-the-box"? :) It's a 3rd party thing... It works well but maybe we should formulate this differently if that's what we're going for.

VALKEY-K8S-OPERATOR
```diff
@@ -0,0 +1,267 @@
...
+- Deployment approach: The operator will rely on either:
+  - A fully open-source Helm chart maintained internally
+  - Direct deployment manifests managed by the operator itself
+
+Comparative notes:
+- SAP’s internal `valkey-operator` is optimized for Sentinel and metrics but is not distribution-friendly.
+- Hyperspike’s operator is more lightweight and public, but lacks Sentinel and TLS support.
+This RFC proposes an open-source Valkey operator that balances ease of use and production-grade features.
+
+Specification
```
zuiderkwast: Before the specification (which I think is more about exactly how to do things) let's gather *requirements* folks have for the different setups. We don't need to implement everything in the first version, but it's good to collect what we already know we need here.

For example, we have some requirements (or needs, wishes, etc) on the operation of a cluster:

```suggestion
## Requirements

### Cluster deployment

* Orchestrate starting up a cluster.
* Orchestrate taking down a cluster.
* Orchestrate rolling upgrades without downtime, using the rolling upgrade procedure described in the Cluster Tutorial: https://valkey.io/topics/cluster-tutorial/#upgrade-nodes-in-a-valkey-cluster
* Orchestrate scale-out and scale-in by adding more nodes (primaries and replicas) and moving slots to them, or away from them, without downtime.
* When adding nodes to an existing cluster:
  - Create a pod and add the node to the cluster using CLUSTER MEET
  - K8s rediness-probe after CLUSTER MEET
  - Check that the node is in the cluster (CLUSTER NODES or similar) and that all nodes in the cluster converge on the same results, i.e. that all nodes in the cluster is aware of the new node before migrating slots to it or making it replicate another node.
  - Rediness gate (on the pod) until a replica has finished loading/replicating.
* Anti-affinity: Primary and its replicas on different workers. Spread primaries
  evenly accross workers.

Specification
```

VALKEY-K8S-OPERATOR.md
```diff
@@ -0,0 +1,267 @@
...
+```yaml
+apiVersion: cache.percona.com/v1alpha1
+kind: Valkey
+metadata:
+  name: my-valkey
+spec:
+  mode: cluster
+  replicas: 3
+  sentinel:
+    enabled: true
```
zuiderkwast, line 119: Cluster mode with sentinel is not possible.

I don't think we should have this yaml code under specification. It looks more like an example to me.

VALKEY-K8S-OPERATOR
```diff
@@ -0,0 +1,267 @@
...
+- Sensitive data like passwords are not stored in the config map and are auto-generated and stored in Secrets.
+
+Keyspace notifications
+
+N/A – this Operator does not modify Valkey internals; only orchestrates deployment.
+
+Cluster mode
+
+- Sentinel mode uses quorum-based failover.
+- Standalone mode supports scaling replicas.
```
zuiderkwast: Sentinel and standalone are not cluster mode.

IIUC, we want three different deployment styles:

1. Standalone (a single node, or with replicas but no automatic failover)
2. Sentinel
3. Cluster

Ah, I see "Custer mode" is a heading from the template. Let's ignore the template. :)

VALKEY-K8S-OPERATOR
```diff
@@ -0,0 +1,267 @@
...
+- Sentinel handles failover logic.
+- Operator ensures correct replica configurations based on `spec.replicas`.
+
+Dependencies
+
+- Bitnami Valkey Helm Chart
+- cert-manager (optional)
+- Prometheus Operator (optional)
+
+Benchmarking
```
zuiderkwast: These headings from the template doesn't make sense for an operator. They are more intended for an RFC describing a new feature in Valkey itself, a feature that extends the module API for example. Let's ignore much of these sections of template and look at the README instead to see what's expected from an RFC.

zuiderkwast (commented):
Added some comments and some of our requirements as suggestions in the comments.

VALKEY-K8S-OPERATOR.md
```diff
@@ -0,0 +1,267 @@
...
+
+This Operator addresses these challenges by:
+- Providing a Kubernetes-native way to deploy and manage cluster mode Valkey, which is the only deployment type that provides HA and horizontal scaling.
+- Enabling self-service for developers via CRDs
+- Automating TLS setup, monitoring, and failover
+- Supporting advanced production use cases:
+  - Cluster mode deployments with multiple Sentinel-managed shards
+  - Persistent storage and backup/restore workflows
+- Providing simplified standalone or Sentinel-only deployments for smaller or test environments where full cluster mode is unnecessary
+- Matching enterprise expectations for security, observability, and compliance
```
bjosv, line 26: - Aiming for Level 5 in Operator SDKs defined [capability levels](https://sdk.operatorframework.io/docs/overview/operator-capabilities/).

VALKEY-K8S-OPERATOR
bjosv: * Support K8s worker node cordon and drain.
* Handle TLS cert reload (`CONFIG SET`).
* Handle etcd failures/restarts & API server unavailability.
* Handle network partitioning of the operator, or primary, or replica.
* JSON structured logging (Alarm/events?)
* Never get stuck in a non-working state.

andrey-glazkov (commented):
adding some simple changes here re Cluster mode deployments and Requirements section

VALKEY-K8S-OPERATOR.md
andrey-glazkov: Start with L3, move from 3 to 5 in increments

VALKEY-K8S-OPERATOR
andrey-glazkov: Andrey to work on this - add a line/two on this

VALKEY-K8S-OPERATOR.md
andrey-glazkov: think about it - maybe some clarity on this

VALKEY-K8S-OPERATOR
jdheyburn: To clarify:

- I am in favour of providing a Helm chart to deploy out the Valkey Operator to Kubernetes
- I am not in favour of the Operator itself wrapping a Helm chart to deploy out resources - my opinion is that the Operator should manage the resource creation itself

VALKEY-K8S-OPERATOR
andrey-glazkov: Live config settings
Module support

VALKEY-K8S-OPERATOR.md
andrey-glazkov: Make it clear that it is an example

VALKEY-K8S-OPERATOR.md
andrey-glazkov: make clear it is an example

Preisschild:
Would it make sense to fork Hyperspike's operator (https://github.com/hyperspike/valkey-operator) into this org and then continue development from there?


It seems that the basics (cluster mode) are already supported

VALKEY-K8S-OPERATOR.md
bjosv: Not `VALKEY-K8S-OPERATOR.md` ?

andrey-glazkov: oops
fixed

VALKEY-K8S-OPERATOR.md
```diff
@@ -0,0 +1,219 @@
+RFC: 21
```
zuiderkwast, line 1: The RFC number is the PR number, which is #28

```suggestion
RFC: 28
```

PR #34, opened 2026-02-03
Valkey support for Top Keys analysis

ranshid:


ValkeyTopKeys.md
```diff
@@ -0,0 +1,242 @@
...
+
+---
+
+## Motivation
+
+The goal is to provide operators with **real-time ability to identify keys that are primary contributors to system load**, 
+allowing them to perform targeted mitigations such as key deletion, redistribute slots or scaling.  
+
+Some examples where a specific key can contribute to resource consumption include:  
+1. Extremely large hash tables can generate large network spikes when commands like `HGETALL` are executed.
```
madolson, line 24: This isn't a problem, this can be easily attributes to command log.

ValkeyTopKeys.md
```diff
@@ -0,0 +1,242 @@
...
+---
+
+## Motivation
+
+The goal is to provide operators with **real-time ability to identify keys that are primary contributors to system load**, 
+allowing them to perform targeted mitigations such as key deletion, redistribute slots or scaling.  
+
+Some examples where a specific key can contribute to resource consumption include:  
+1. Extremely large hash tables can generate large network spikes when commands like `HGETALL` are executed.  
+2. Very large sets can cause extended server unresponsiveness when executing commands such as `SDIFFSTORE`.
```
madolson, line 25: Same, can easily be identified by command log

ValkeyTopKeys.md
```diff
@@ -0,0 +1,242 @@
...
+
+We cannot require the user to plan for the specific debug case he did not expect, Since these are mostly used in order to debug UNEXPECTED events.
+For this reason, activating the feature should not be a blocker to identify an already-ongoing situation.
+Fine tuning configurations, can be considered, but also cannot become a blocker to identify an already ongoing issue.
+
+---
+
+### 3. Integrability
+
+- Output MUST be suitable for aggregation into cluster-wide or database-wide views.
```
madolson, line 78: We already don't have database wide views, why does this need to be database wide?

ValkeyTopKeys.md
```diff
@@ -0,0 +1,242 @@
...
+   2) (float) <qps>
+3) ...
+```
+---
+
+### TOPKEYS
+
+Returns Top-N keys by size characteristics.
+```
+TOPKEYS <CARD | MEMORY> TOP <N>
```
madolson, line 143: You mention DB awareness, these requests are not per-DB.

ValkeyTopKeys.md
```diff
@@ -0,0 +1,242 @@
...
+
+---
+
+### Configuration (Optional)
+
+These are high-level suggested configurations. Additional tuning options may be added later.
+
+### Hot Keys
+
+`hotkeys-max-n <integer>`
```
madolson: Unclear why this needs to be a config, could just be part of the API.

ValkeyTopKeys.md
```diff
@@ -0,0 +1,242 @@
...
+1. **Write Access Frequency**
+   - Refers to the access rate of a key by a WRITE command.
+   - Number of write operations per key over a configurable time window.
+2. **Read Access Frequency**
+   - Refers to the access rate of a key by a READ command.
+   - Number of read operations per key over a configurable time window.
+3. **Key Cardinality**
+   - Refers to the cardinality of the key. This will be the number of bytes in a string type, number of items in a hash type, number of items in a stream type etc...
+4. **Key Memory Usage**
+   - Refer to the amount of memory consumed by a key. This should be identical to the output of the command `MEMORY USAGE <key>`
```
madolson, line 119: If we have memory usage, why do we need cardinality? It feels like memory is strictly more useful.

ValkeyTopKeys.md
```diff
@@ -0,0 +1,242 @@
...
+### Hot Keys
+
+`hotkeys-max-n <integer>` 
+Maximum number of keys returned for HOTKEYS.
+Controls memory usage and ensures Top-N lists remain bounded.
+Default: 10
+
+`hotkeys-read-access-threshold <integer>` - default 3000
+`hotkeys-write-access-threshold <integer>` - default 2000
+Threashold configuration.  Only keys exceeding these QPS thresholds appear in HOTKEYS output. Prevents low-activity keys from cluttering results.
```
madolson: I don't understand these configs, and more broadly how hot keys will work. Is it the current hot keys (All current keys accessed more than 3000 times) is it keys that were hot at some point (sort of like slowlog, a given key accessed more than X times in the past).

ValkeyTopKeys.md
ranshid: I agree. and I stated later that COMMAND LOG is a valid solution for some cases (like this).
Still, I think some users would love to get a more generic way to analyze the KEYs which are the root cause for different issues they get instead of cross-analyzing different statistics.

ValkeyTopKeys.md
ranshid: agree. same response as before. command log is a fine alternative. For root causing issues, I agree command log might be enough for most cases. I do think that in some cases users also want to understand the potential issues they might experience doing some "database analysis" in order to identify what is the largest keys they use without going to understanding this from they application side. This is not RCA, and maybe I should add this to the motivation section?

ValkeyTopKeys.md
ranshid: Yeh. I kinda battled my thoughts on how should we correctly expose these statistics. TBH I think that in most cases application would like a complete dataset analysis (not only per specific database).

ValkeyTopKeys.md
ranshid: right. they require one to select the DB first. (or we can add the db as an optional argument)

ValkeyTopKeys.md
ranshid: TBH we can. But I think this will kinda force the implementation to be less frugal in resources like memory and CPU.

ValkeyTopKeys.md
ranshid: Agree. I think we can decide on only one. but then we need to decide if memory is worth implementing the tracking investment. I mean with valkey-cli I think users are always using bigkeys analysis and not memkeys. this is probably since memkeys is so much more expensive in CPU and time.

ValkeyTopKeys.md
ranshid: Sure. I can explain more but I did not want to go into the hotkey algorithm here, as it is being discussed in an already existing PR. I think maybe I will remove these configs from the proposal in the RFC and we can discuss specific configs as part of the detailed PR.

ranshid:
@madolson Thank you for taking the time to review!
I know this is very raw and needs several iterations to focus the discussion on the major decisions, but wanted to start somewhere.
I made some changes following your comments. we can circle more.

PR #35, opened 2026-02-11
Modular Clustering

murphyjacob4:
This RFC is part 1 of 2 for "Cluster V2" https://github.com/valkey-io/valkey/issues/384

1) [This Proposal] Introduce a modular Core <-> Clustering interface
2) Introduce a first-party cluster implementation following from 1

### DIFF
--- a/ModularClustering.md
+++ b/ModularClustering.md
@@ -1,0 +1,1106 @@
+---
+RFC: 35
+Status: Proposed
+---
+
+# Valkey Modular Clustering RFC
+
+## Abstract
+
+This RFC proposes a modular interface to decouple Valkey's cluster management
+from its core execution engine, allowing for pluggable clustering
+implementations via **Cluster Provider Modules**. Currently, Valkey’s clustering
+logic is tightly coupled to the legacy Redis Cluster specification, preventing
+the adoption of modern consensus algorithms.
+
+This proposal defines a minimal API where the Module manages topology and
+consistency while the Core maintains control over business logic and data
+routing. Key architectural contributions include:
+
+- Abstracted Membership: A push-based model where the Module informs the Core of
+  topology changes, enabling diverse backends (e.g., Raft, Multi-Raft, or
+  external stores like etcd).
+- Opaque Metadata Layer: Two flexible key-value stores (one global and eventual,
+  one shard-level and durable) allowing cluster state transitions without
+  understanding the underlying implementation.
+- Module Heartbeats: Modules send heartbeats to the Core periodically to inform
+  that the cluster is in "OK" state. The Core does not interpret the underlying
+  consensus state; it simply consumes the topology updates provided by the
+  Module.
+
+This interface aims to solidify the interaction between the Core and Clustering,
+allowing both to evolve independently.
+
+## Motivation
+
+We’ve previously made an attempt to refactor clustering to support new
+clustering systems in Valkey by refactoring cluster.c/h into cluster_legacy.c/h
+and cluster.c/h. However, cluster.c is heavily coupled to the type of
+implementation that cluster_legacy.c provides.
+
+This proposal seeks to define a general purpose, minimal, clustering interface
+allowing pluggable clustering implementations via module. Valkey cluster then
+would operate in one of two modes:
+
+- Cluster v1 - the existing implementation, which is used if no clustering
+  module is supplied
+- Module clustering - an arbitrary implementation of clustering that provides
+  membership and metadata management.
+
+The goal is not to reimplement cluster v1 as a module - this module API makes no
+effort to support such a refactor. Cluster v1 requires the clustering to have a
+deep understanding into all of Valkey's clustering metadata. The module-based
+clustering interface is intended to provide a cleaner abstraction that is mostly
+opaque to how the Valkey Core will use it. The goal is to provide a complete
+enough API that the Core can build off of, without coupling the two components
+and preventing independent growth.
+
+Later, we can reuse this module-based system to ship new built-in clustering
+modes (e.g. over Raft). Such implementations could use the same interfaces while
+being statically loaded inside the Core itself.
+
+To help with reading this proposal, the following architectures will be used as
+potential cluster provider module implementations:
+
+- **Cluster-Level Raft**: A clustering implementation that uses a global Raft
+  group consisting of all cluster members.
+- **Multi-Raft (Shard-Level Raft)**: A clustering implementation consisting of
+  many per-shard Raft groups.
+- **External (e.g. Etcd/Zookeeper)**: A clustering implementation that utilizes
+  an external store for cluster management.
+
+## Specification
+
+## 1. Cluster Module Registration
+
+Registration occurs through a module API:
+
+```
+/* Register your module as a cluster provider. This must be done in the
+ * VM_ModuleLoad callback. May return an error if some other cluster provider is
+ * registered (at most one provider can register).
+int VM_RegisterClusterProvider(ValkeyModuleCtx *ctx, ClusterOps *ops)
+```
+
+For simplicity, Cluster Provider Modules will not support `MODULE UNLOAD`. The
+Core will deny `MODULE UNLOAD` of a registered Cluster Provider.
+
+### 1.1 ClusterOps Struct
+
+ClusterOps is a versioned struct containing all the functions for Core -> Module
+communication:
+
+```
+typedef void(*JoinClusterCallback)(int success, ValkeyModuleString *err);
+typedef void(*RemoveFromClusterCallback)(int success, ValkeyModuleString *err);
+typedef void(*JoinShardCallback)(int success, ValkeyModuleString *err);
+typedef void(*RemoveFromShardCallback)(int success, ValkeyModuleString *err);
+
+typedef struct ClusterOps {
+   /* Struct version, for backwards compatibility */
+   int version;
+
+   /* ---------------------------- Membership ------------------------------ */
+   /* Join into a cluster given some opaque endpoint information. E.g. CLUSTER
+    * MEET <argv>. <argv> can be changed based on implementation.
+    *
+    * After joining or failing to join, the module should trigger the callback
+    * with details. */
+   int (*JoinCluster)(ValkeyModuleCtx *ctx, ValkeyModuleString *argv,
+                      JoinClusterCallback callback);
+
+
+   /* Remove the target node from my cluster. E.g. CLUSTER FORGET node-id. If
+    * node-id is myself, gracefully remove my cluster membership.
+    *
+    * After removing or failing to remove, the module should trigger the
+    * callback with details. */
+   int (*RemoveFromCluster)(ValkeyModuleCtx *ctx, ValkeyModuleString *node_id,
+                            RemoveFromClusterCallback callback);
+
+   /* Join into a given shard within the cluster. E.g. CLUSTER REPLICATE.
+    *
+    * After joining or failing to join, the module should trigger the callback
+    * with details. */
+   int (*JoinShard)(ValkeyModuleCtx *ctx, ValkeyModuleString *shard_id,
+                    JoinShardCallback callback);
+
+
+   /* Remove the given node from my shard. If the node is myself, gracefully
+    * remove my shard membership. E.g. CLUSTER REPLICATE NO ONE.
+    *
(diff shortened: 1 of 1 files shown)


ModularClustering.md
```diff
@@ -0,0 +1,1106 @@
...
+    * per-shard Log. */
+   struct ValkeyModuleClusterMetadataOperation {
+      /* The key to update. If it doesn't exist, will be inserted. */
+      ValkeyModuleString *key;
+      /* The new value of the metadata */
+      ValkeyModuleString *value;
+      /* The expected current revision. If not the actual latest, the Module
+       * must fail the operation. If the key does not currently exist, use
+       * NULL. This provides compare-and-swap (CAS) semantics. */
+      void *expected_revision;
```
allenss-amazon, line 149: Any reason we can't be more specific about the data type of "version"? If it's intended to be an opaque type (defined by the cluster provider), then it should have it's own type, with a vtable, not void.

ModularClustering.md
```diff
@@ -0,0 +1,1106 @@
...
+   }
+
+   enum ValkeyModuleGossipPropagation {
+     /* Gossip to all shard members. */
+     SHARD = 1,
+     /* Gossip to all cluster members. */
+     GLOBAL = 2
+   }
```
allenss-amazon: Is the gossip parameter just a hint about distribution OR does it imply that the associated key is actually shard-local in it's value rather than global? If the later, I'd rather see the the scope somehow bound to the name itself so that there's no accidental reuse of names, etc. In other words, could there be a global namespace and a shard-local namespace? And if I have that, does it mean that this gossip parameter is no longer explicit, but implied by the name?

ModularClustering.md
```diff
@@ -0,0 +1,1106 @@
...
+
+Revisions are specified as `void*` pointers. Revisions are either:
+
+- `NULL` - if no previous revision of the metadata exists (initial insert)
+- A previously returned revision - if the engine is aware of a previous revision
+  (update)
+
+`void*` pointers are used to allow flexibility in how revisions are defined. The
+details of this will be left to the Module (e.g. in Raft, it may include both
+term and index).
```
allenss-amazon, line 325: Nit: rather than void, it should be an opaque type that has a vtable exported from the module.

ModularClustering.md
```diff
@@ -0,0 +1,1106 @@
...
+
+- `SHARD`: best effort attempt to propagate to current shard members (e.g.
+  replication offset)
+- `GLOBAL`: best effort attempt to propagate to all cluster members (e.g. slot
+  routing, shard primaryship)
+
+Note: propagation is expressed as minimum guarantees. Modules are free to
+propagate beyond the specified scope (e.g. propagate SHARD to all nodes). For
+that reason, the Core should ensure shard-gossipped metadata is globally unique
+(e.g. by prefixing with node ID or shard ID).
```
allenss-amazon, line 377: This feels like you're really defining a third type of data, Why not simply define a shard-local eventually consistent namespace?

ModularClustering.md
```diff
@@ -0,0 +1,1106 @@
...
+is still alive.
+
+To handle 2), the Module should implement periodic heartbeats from the Core main
+thread. They can do this using a number of methods (main thread timer callbacks,
+event loop event registration, local PING requests to the Valkey port). It is
+left to the modules to design this system.
+
+### 4.2 Remote Failure
+
+When the Module detects a remote failure, the Module is expected to callback via
```
allenss-amazon, line 444: What is the responsibility within the module for detecting remote failures? Is this time-bounded? Is the intra-shard timeframe the same as the cross-shard timeframe?

ModularClustering.md
```diff
@@ -0,0 +1,1106 @@
...
+control over slot distribution.
+
+Slot assignment is broken into two metadata pieces: one in the Gossip and one in
+the Shard Log:
+
+- The Gossip metadata we will refer to as the "Slot Routing Table" and is
+  gossipped throughout the cluster for the purposes of routing.
+- The Shard Log metadata we will refer to as the "Shard Slot Ownership" and is
+  used to ensure the slot assignment is durable through failovers at the shard
+  level.
```
allenss-amazon: Who is responsible for handling conflicts and reconciliation between the shard log and the gossip?

ModularClustering.md
```diff
@@ -0,0 +1,1106 @@
...
+data ownership concept. The Valkey Primaryship is just stored as opaque
+metadata.
+
+When failovers happen, modules may choose to implement hooks
+(`VALKEYMODULE_EVENT_REPLICATION_ROLE_CHANGED` server event) to perform the
+necessary failover in their consensus layer to align these two concepts. This
+choice is left up to the Module and is not strictly required, but may make
+implementation easier.
+
+## 5 Cluster Formation and Operation Details
```
allenss-amazon, line 477: It's not entirely clear how much of the code in this section is in the module and how much is built on top of the module's API.  I think the intention is mostly the later -- right?

ModularClustering.md
```diff
@@ -0,0 +1,1106 @@
...
+logic is tightly coupled to the legacy Redis Cluster specification, preventing
+the adoption of modern consensus algorithms.
+
+This proposal defines a minimal API where the Module manages topology and
+consistency while the Core maintains control over business logic and data
+routing. Key architectural contributions include:
+
+- Abstracted Membership: A push-based model where the Module informs the Core of
+  topology changes, enabling diverse backends (e.g., Raft, Multi-Raft, or
+  external stores like etcd).
```
madolson, line 22: The current legacy split was to have the module maintain it's own copy of the topology, that the engine could fetch values from. The same is roughly true of the next two points, the engine is just getting the data that it wants. It requires a lot of boiler plate code in the theoretical future module, but is not dictating the implementation. This new proposal is also tight coupling with a specific backend.

ModularClustering.md
```diff
@@ -0,0 +1,1106 @@
...
+
+When the Module detects a remote failure, the Module is expected to callback via
+`VM_ClusterProviderUpdateMembership`. Modules have the option of marking nodes
+as either:
+
+- `OK` - node is online, or began failing recently enough that it is yet to be
+  marked unhealthy
+- `PFAIL` - node is not yet canonically unhealthy (by the Module’s definition)
+  but is locally unhealthy. It is not strictly needed, but will be client
+  visible as PFAIL flag in the cluster topology
```
madolson, line 452: We should kill off pfail. It's not practically useful to end clients.

ModularClustering.md
```diff
@@ -0,0 +1,1106 @@
...
+manually (e.g upon partition detection):
+
+```
+/* Immediately enter cluster down state, ignoring any remaining TTL */
+void VM_ClusterProviderClusterDown()
+```
+
+Many modules will use the heartbeat to implement a leasing system in order to
+detect network partitions. But it also serves to inform the Core that the Module
+is still alive.
```
madolson, line 435: It seems odd the module is down? What exactly does the module being alive mean? The core can be down because of long running, but we should able to better fence of the module.

madolson (commented):
So, the main comment is just that I don't know if this is really better than the direction of the previous cluster V2. It was intended to require more implementation, but also meant that the backend could do literally whatever it wanted to do to materialize the clustering state. This new implementation seems to force a specific architecture on the backends, which seems less flexible overall to me.

murphyjacob4:
> So, the main comment is just that I don't know if this is really better than the direction of the previous cluster V2.

Is there somewhere the proposal for previous cluster v2 is documented/proposed? It's hard to discuss the merits of it since I don't have a very concrete idea of what all it proposed. It iterated a lot so I am also not sure which version you are referring to

That said - from what I can tell - no iteration of cluster V2 made an attempt to define a clear core <-> clustering contract. With the proposed split in this RFC, I think it would be really easy to do things like: Active-Active (store in shard-log metadata), Synchronous Replication (store in shard-log metadata), Cluster-level configs (store in gossip), Cross-Cluster replication (basically just a cluster-level config on the target cluster, then each shard follows the source shard). The primitives feel strong enough that the core would be less burdened with reasoning about the distributed state.

> This new implementation seems to force a specific architecture on the backends, which seems less flexible overall to me.

So my whole intention was to decouple from architecture specifics. Maybe you can provide an example of an implementation that would be hard to implement on the new model? I provided examples throughout of a few different backends and how they would implement this but maybe there is some other way that you are thinking?

ModularClustering.md
murphyjacob4: The only place I feel like I remember using PFAIL was in coordinated searches in ValkeySearch, we try to avoid sending to nodes in the shard that are in PFAIL. For that reason, I felt like it could be useful

ModularClustering.md
murphyjacob4: If the module is on the main thread, then yeah this doesn't really make sense. But I figured it may make sense to do consensus work on a background thread. In which case you can be exposed to things like deadlock or busy loops which prevent the heartbeat from arriving. From the core's point of view, it wouldn't be any different from a partition, and it shouldn't violate the correctness

ModularClustering.md
murphyjacob4: > The current legacy split was to have the module maintain it's own copy of the topology, that the engine could fetch values from

Yeah I don't like how we refer to "topology" since it is really a mix of membership (who is in the cluster, which shard are they in) and metadata (which shards own what slots, which node leads a shard). To me that feels like a leaky abstraction.

The goal here was to really tease out those details so we don't end up with a vague concept of "topology". Now you have membership (static, transparent to module) and metadata (dynamic, opaque to the module). If, for a dumb/crazy example, tomorrow we wanted to change the number of slots to 32k, the clustering layer does not need to be aware of that. I think this is the right abstraction to make, but it breaks from cluster v1's "slots are transparent" and that's why I don't think we should bother with coming up with one uber-design that encompasses both.

The above goes for primaryship as well, although this one has more concrete examples. We will need to support more complex concepts there, e.g. active-active, synchronous replication, and there is no reason to force the module to handle those cases IMO.

All that said, we could say slots are a firm contract and make it visible to the clustering layer. Same for shard ownership/primaryship. There isn't necessarily anything really wrong with that, but it is just a more strict contract and I don't really know why we would need that if we can make the abstraction higher level.

murphyjacob4: > This new proposal is also tight coupling with a specific backend.

You are going to need to explain more, is it just because we split Gossip and Shard Log? There are other ways we can model that (e.g. propagation concern [shard/cluster/etc] and write concern [shard quorum/not]) but I didn't want to explode the interface into a bunch of permutations

ModularClustering.md
murphyjacob4: I'm open to it. I iterated a lot on this. E.g. another idea was one Metadata channel with configurable write/propagation concern. E.g. `UpdateMetadata(propagation=SHARD/GLOBAL, write_concern=LOCAL/SHARD_QUORUM, ...)` but you end up with some nonsense options like `prop=GLOBAL, write_concern=SHARD_QUORUM` which I didn't like

So I guess we could have something like three channels:

1. Gossip-only, shard-local
2. Gossip-only, global
3. Quorum-commit, shard-local

ModularClustering.md
murphyjacob4: Yeah it is arbitrary, e.g. in raft I believe you need to specify term and index. I'm also fine defining it as a type. I figured when it comes to implementation some of the API might change a little (based on what is most ergonomic). Hard to anticipate now

ModularClustering.md
murphyjacob4: I didn't define this since it will depend on how the clustering implementation aims to guarantee correctness. Modules interested in correctness/write consistency will need to implement the primary health via leasing/heartbeats. So like for Raft, primary needs to append an entry every X seconds, and when replicas don't see it for X*2 seconds, they need to mark it down. In theory a module could also have a less-correct implementation (a la cluster v1). If we go with this design and we implement our own module, it will come with a separate design that hashes these details out.

ModularClustering.md
murphyjacob4: Yeah I tried to split it like so:

* Membership: code would mostly be in the Module. The Core just tells it what modifications to make and it makes those changes opaque to the core (e.g. following from Raft membership changes spec)
* Metadata: code would be mostly in the Core. The Core will tell the module to change metadata in some way, but the module does not know why. The Module is just responsible for the consistency/durability/propagation of that opaque metadata

I can split the examples out to better show this

